Issue #37, opened 2025-02-03
`withSnapWP()` causes fatal error on Windows


justlevine:
### Description

When running `npm run dev` or `npm run build` on a Windows machine (terminal, gitbash, cmd) I get the following error:

```
 ⨯ Failed to load next.config.mjs, see more info here https://nextjs.org

/docs/messages/next-config-error

> Build error occurred
[Error: Only URLs with a scheme in: file, data, and node are supported by the default ESM loader. On Windows, absolute paths must be valid file:// URLs. Received protocol 'c:'] {
  code: 'ERR_UNSUPPORTED_ESM_URL_SCHEME'
}
```

Removing `withSnapWP()` from `next.config.mjs` allows `dev`/`build` to finish successfully, so my assuming is this is due to how the file is trying to locate the `snapwp.config.mjs`

### Link to the code that reproduces this issue

https://github.com/rtCamp/snapwp/blob/snapwp%400.0.3/examples/nextjs/starter/next.config.mjs#L3

### To Reproduce

1. On _Windows_, use `npx snapwp` (or clone the repo and manually update the `starter/.env`
2. Run `npx next dev` (to run w.out our concurrency) or `npm run build`
3. See error

4. replace `export default withSnapWP({});` with `export default {}` 
5. run `npm run dev` or `npm run build`, and see the error is gone.

### Additional context

![Image](https://github.com/user-attachments/assets/7b42ad3c-ea83-4bef-be71-165ef8d67153)

### Frontend Environment

```bash
Operating System:
  Platform: win32
  Arch: x64
  Version: Windows 11 Pro
  Available memory (MB): 32453
  Available CPU cores: 24
Binaries:
  Node: 20.11.1
  npm: 10.2.4
  Yarn: 1.22.22
  pnpm: N/A
Relevant Packages:
  next: 15.1.6 // Latest available version is detected (15.1.6).
  eslint-config-next: 14.2.5
  react: 19.0.0
  react-dom: 19.0.0
  typescript: 5.3.3
Next.js Config:
  output: N/A
```

### Backend Environment

```bash

```

### Please confirm that you have searched existing issues in the repo.

- [x] Yes

SH4LIN:
Hey, @justlevine, I'll look into this.

PR #39, opened 2025-02-04
Fix: Absolute Path import issue with windows - ERR_UNSUPPORTED_ESM_URL_SCHEME

SH4LIN:
## What

- This PR fixes the `ERR_UNSUPPORTED_ESM_URL_SCHEME` error, which occurs on Windows when using an absolute path without the `file:` or `data:` scheme.

## Why

- On Windows, `C:/` is interpreted as a protocol, making it an invalid protocol. Since it represents a directory, absolute paths in the Windows filesystem require a scheme to function correctly.

### Related Issue(s):

- Closes: #37 

## How

- First, we normalize the path using the `path` Node.js API.  
- Then, we convert it to a file URL using the `url` Node.js API.

## Testing Instructions

1. Scaffold the example in your Windows system.
2. Install all dependencies
3. Execute npm run dev or npm run build - It should be executed successfully.

## Checklist

### DIFF
--- a/packages/next/src/config/with-snap-wp.ts
+++ b/packages/next/src/config/with-snap-wp.ts
@@ -83,24 +83,23 @@ const withSnapWP = async ( nextConfig?: NextConfig ): Promise< NextConfig > => {
 	];
 
 	// Locate the SnapWP configuration file.
-	let snapWPConfigPath = possibleSnapWPConfigPaths.find( ( path ) => {
-		return fs.existsSync( path );
-	} );
+	let snapWPConfigPath = possibleSnapWPConfigPaths.find(
+		( possibleSnapWPConfigPath ) => {
+			return fs.existsSync( possibleSnapWPConfigPath );
+		}
+	);
 
 	if ( ! snapWPConfigPath ) {
 		throw new Error( 'SnapWP configuration file not found.' );
 	}
 
 	// Platform-specific handling
-	if (process.platform === 'win32') {
+	if ( process.platform === 'win32' ) {
 		// Use path.normalize to replace backslashes correctly
-		snapWPConfigPath = path.normalize(snapWPConfigPath);
+		snapWPConfigPath = path.normalize( snapWPConfigPath );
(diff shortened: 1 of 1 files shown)


packages/next/src/config/with-snap-wp.ts
```diff
@@ -81,14 +83,24 @@ const withSnapWP = async ( nextConfig?: NextConfig ): Promise< NextConfig > => {
...
 	if ( ! snapWPConfigPath ) {
 		throw new Error( 'SnapWP configuration file not found.' );
 	}
 
+	// Platform-specific handling
+	if ( process.platform === 'win32' ) {
+		// Use path.normalize to replace backslashes correctly
+		snapWPConfigPath = path.normalize( snapWPConfigPath );
+		// Convert it to a file:// URL
+		snapWPConfigPath = url.pathToFileURL( snapWPConfigPath ).href;
```
justlevine, line 101: Is there a reason to have this inside (vs) outside the conditional?

SH4LIN: we need a URL like `file://C:/xyz.mjs` only for windows (win32)

justlevine: But we _can_ use it everywhere, right?

Ideally we want as much as our code to be cross-env as possible.

SH4LIN: > But we can use it everywhere, right?

I have tested the code `snapWPConfigPath = url.pathToFileURL( snapWPConfigPath ).href;` outside the if condition on windows and on mac, working as expected. Not sure about linux I don't have linux system.

SH4LIN: It should technically work, but I can't say for sure without testing it.

justlevine: It's a nitpick so there's no point sending you back into the codebase for this.

justlevine (approved):
Working locally but I didnt stress test, since #41 is up next.

PR #42, opened 2025-02-05
feat: migrate block definitions to Config API

Swanand01:
## What

This PR moves the custom block definitions to the `snapwp.config.ts` file.
Instead of passing them as props to the `EditorBlockRenderer`, they are now imported from `getConfig`

## Why

Since block definitions are overloadable/configurable, they should be put in the `snapwp.config.ts` file.

### Related Issue(s):

https://github.com/orgs/rtCamp/projects/141/views/14?pane=issue&itemId=96524753&issue=rtCamp%7Cheadless%7C396
https://github.com/rtCamp/headless/issues/396

## How

The `blockDefinitions` property is added to the `SnapWPConfig` interface and `SnapWPConfigManager.snapWPConfigSchema`

## Testing Instructions

1. Create a file named `core-quote.tsx` in `examples/nextjs/starter/src/app`.
2. Override the block implementation in this file
3. In `snapwp.config.ts`, add:
```ts
import CoreQuote from './src/app/core-quote';
import { BlockDefinitions } from '@snapwp/types';

const blockDefinitions: BlockDefinitions = {
	CoreQuote: CoreQuote,
};

const config: SnapWPConfig = {
	blockDefinitions,
};

export default config;
```
4. Add the Quote block to a page, and open the page from the next app.
5. You should see the overridden implementation.

## Checklist


Swanand01:
@ayushnirwal @justlevine What kind of validation should we add here:
https://github.com/rtCamp/snapwp/blob/be538aec0ff82d4e3a8301c92780e291d6ef8eb0/packages/core/src/config/snapwp-config-manager.ts#L200-L203

Swanand01:
@justlevine Do we want to enforce lazy loading for the custom block implementations, or do we leave it up to the user to use `React.lazy` / normal imports?

On a side note, I haven't been able to verify if using `React.lazy` while passing block definitions is working as intended. Have discussed the same with @ayushnirwal over a call.

Swanand01:
This is blocked because we cannot import .tsx block definition files into snapwp.config.mjs.
The PR will be unblocked when we move snapwp.config to .ts

cc: @BhumikP

BhumikP (approved):
Tested the changes, LGTM

ayushnirwal (approved):

ayushnirwal:
@Swanand01 I have updated the issue with the additional tasks discussed.

packages/blocks/src/editor-blocks-renderer/index.tsx
```diff
@@ -10,14 +12,13 @@ type EditorBlocksRendererProps = {
...
  * A react component to render editor blocks.
  * @param props - Props.
- * @param props.blockDefinitions - blocks rendering functions.
  * @param props.editorBlocks - A list of blocks to be rendered.
  * @return The rendered template
  */
 export default function EditorBlocksRenderer( {
 	editorBlocks,
-	blockDefinitions,
 }: EditorBlocksRendererProps ) {
```
justlevine, line 20: Pros vs cons about leaving keeping this prop, and then having a hierarchy of `props?.blockDefinitions || getConfig()?.blockDefinitions || defaultBlockDefinitions`

justlevine (changes requested):
🚀

- Should this PR include tests before merging? Why (not)?

It definitely requires an update our `config-api.md`, and `overloading-wp-blocks.md`, and a changeset (`minor` as is if we're removing the prop, `patch` if we're keeping it - changesets doesnt understand SemVer at <1.0.0)

packages/blocks/src/editor-blocks-renderer/index.tsx
Swanand01: Would you please tell me a case where the user would want to overwrite the block definitions passed from the config file by passing them as props to `EditorBlocksRenderer`?
I can't think of any pros of doing this at the moment.
Also, we are doing something similar to `props?.blockDefinitions || defaultBlockDefinitions` here:
https://github.com/rtCamp/snapwp/blob/806792d153651dc2355ce58876fe251e9cb1de7a/packages/blocks/src/block-manager/index.tsx#L47-L70

examples/nextjs/starter/package.json
```diff
@@ -35,6 +35,7 @@
 		"@snapwp/codegen-config": "file:../../../packages/codegen-config",
 		"@snapwp/eslint-config": "file:../../../packages/eslint-config",
 		"@snapwp/prettier-config": "file:../../../packages/prettier-config",
+		"@snapwp/types": "file:../../../packages/types",
```
justlevine, line 38: I don't believe the example is using this directly. Revert?

.changeset/brave-baths-agree.md
```diff
@@ -0,0 +1,6 @@
+---
+"@snapwp/blocks": patch
+"@snapwp/core": patch
+---
+
+migrate block definitions to Config API
```
justlevine, line 6:
```suggestion
feat: Add support for providing block definitions via the Config API.
```

docs/overloading-wp-behavior.md
```diff
@@ -124,56 +124,114 @@ The `@snapwp/blocks` package provides a large and growing number of [Block Compo
 
 First, identify the block you want to overload. Each block has a unique name (e.g., `CoreParagraph`, `CoreImage`). You can find this name in the block's data when fetched via WPGraphQL.
 
-### 2. Creating a Custom Component
+### 2. Creating a Custom Component for Core Paragraph
```
justlevine, line 127: Can we revert this whole section?

The types and real-world implementation turn the example to noise in context ("Why would I overload a paragraph block to display the cookies).

And clearly the block types need more work before we start telling users to consume something generic like `BaseAttributes` and know that means a block component...

docs/overloading-wp-behavior.md
```diff
@@ -124,56 +124,114 @@ The `@snapwp/blocks` package provides a large and growing number of [Block Compo
...
 
+This allows you to apply the override only on specific routes while using the default block rendering elsewhere.
+
 Now, whenever a `core/paragraph` block is encountered, your `MyCustomParagraph` component will be used to render it. Any other blocks will use the default rendering unless you provide a custom component in `blockDefinitions`.
 
+> [!TIP]
+> If a block is overridden both globally (`snapwp.config.ts`) and per-route (`EditorBlocksRenderer` prop), the per-route override takes precedence.
+
 > [!TIP]
 > Setting the block definition value to `null` will use the [default block rendering](#default-html-rendering-with-parse-and-html-react-parser).
```
justlevine, line 237: In context, this should now move up to the global overrides section, and include a `blockDefinitions: {...}` snippet.

Then, in the per-route example, you can just inline the `CoreSomethingElse: null, // Use html-react-parser instead of a Block Component.` into the existing snippet.

docs/overloading-wp-behavior.md
```diff
@@ -257,6 +315,12 @@ export default function Page() {
 }
 ```
 
+> [!NOTE]
+> Similar to blocks, you can override template parts in two ways:
+
+-   Globally by defining overrides in `snapwp.config.ts`.
+-   Per route by passing `blockDefinitions` as a prop to `EditorBlocksRenderer`.
```
justlevine, line 322: You need to prefix them with `>` if you want them in the call out. 

But also, kinda redundant.

How about something closer to "Template Parts are just another type of Block, which means they can be overwritten `[the same way]( link to blocks section of this doc)` in the existing note at the top of this section?

justlevine (changes requested):
Some feedback on the docs, otherwise LGTM 🙌

justlevine (approved):
@ayushnirwal I know you approved this previously, pinging you for rereview just in case something slipped in since.

ayushnirwal (approved):

PR #45, opened 2025-02-06
refactor: codegen config schema property moved to the codegen config package.

SH4LIN:
## What

- This PR will refactor the common codegen config code to the @snapwp/codegen-config package.

## Why

- This refactoring is to avoid code duplication and keeping the minimal code on the userland

### Related Issue(s):

- Part of: https://github.com/rtCamp/headless/issues/303

## How

- Added withCodegenConfig function which will extend the base config with the custom configs provided from the individual codegen.ts file.

## Checklist

### DIFF
--- a/examples/nextjs/starter/codegen.ts
+++ b/examples/nextjs/starter/codegen.ts
@@ -1,28 +1,13 @@
 import type { CodegenConfig } from '@graphql-codegen/cli';
 import { sync as globSync } from 'glob';
-import baseConfig from '@snapwp/codegen-config';
-import { generateGraphqlUrl } from '@snapwp/core';
+import { withCodegenConfig } from '@snapwp/codegen-config';
 import 'dotenv/config';
 
 const GRAPHQL_GLOB = './src/**/*.graphql';
 const graphqlFiles = globSync( GRAPHQL_GLOB );
 
-const config: CodegenConfig = {
+const config: Partial< CodegenConfig > = {
 	...( graphqlFiles.length > 0 && { documents: GRAPHQL_GLOB } ),
-	...baseConfig,
-	// Use the schema file if it's set by CI.
-	schema: process.env.GRAPHQL_SCHEMA_FILE ?? [
-		{
-			[ generateGraphqlUrl(
-				process.env.NEXT_PUBLIC_WORDPRESS_URL,
-				process.env.NEXT_PUBLIC_GRAPHQL_ENDPOINT
-			) ]: {
-				headers: {
-					Authorization: `${ process.env.INTROSPECTION_TOKEN }`,
-				},
-			},
-		},
-	],
 };
 
-export default config;
+export default withCodegenConfig( config );
--- a/packages/codegen-config/src/index.ts
+++ b/packages/codegen-config/src/index.ts
@@ -1,13 +1,54 @@
 import type { CodegenConfig } from '@graphql-codegen/cli';
+import { generateGraphqlUrl } from '@snapwp/core';
 
-const baseConfig: CodegenConfig = {
-	overwrite: true,
-	generates: {
-		'src/__generated/': {
-			preset: 'client',
-			plugins: [],
+/**
+ * Base configuration for GraphQL Codegen.
+ *
+ * @return The base configuration.
+ */
+const baseConfig = (): CodegenConfig => {
+	return {
+		overwrite: true,
+		generates: {
+			'src/__generated/': {
+				preset: 'client',
+				plugins: [],
+			},
 		},
-	},
+		// Use the schema file if it's set by CI.
+		// eslint-disable-next-line n/no-process-env
+		schema: process.env.GRAPHQL_SCHEMA_FILE ?? [
+			{
+				[ generateGraphqlUrl(
+					// eslint-disable-next-line n/no-process-env
+					process.env.NEXT_PUBLIC_WORDPRESS_URL,
+					// eslint-disable-next-line n/no-process-env
+					process.env.NEXT_PUBLIC_GRAPHQL_ENDPOINT
+				) ]: {
+					headers: {
+						// eslint-disable-next-line n/no-process-env
+						Authorization: `${ process.env.INTROSPECTION_TOKEN }`,
+					},
+				},
+			},
+		],
+	};
 };
 
-export default baseConfig;
+/**
+ * Extend the base configuration with the provided configuration.
+ *
+ * @param config The configuration to extend the base configuration with.
+ *
+ * @return The extended configuration.
+ */
+const withCodegenConfig = (
+	config: Partial< CodegenConfig >
+): CodegenConfig => {
+	return {
+		...baseConfig(),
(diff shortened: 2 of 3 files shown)


examples/nextjs/starter/codegen.ts
```diff
@@ -1,28 +1,13 @@
...
-				headers: {
-					Authorization: `${ process.env.INTROSPECTION_TOKEN }`,
-				},
-			},
-		},
-	],
 };
 
-export default config;
+export default withCodegenConfig( config );
```
justlevine, line 13: What does a wrapper function give us that object destructuring doesnt?

packages/codegen-config/src/index.ts
```diff
@@ -1,13 +1,54 @@
 import type { CodegenConfig } from '@graphql-codegen/cli';
+import { generateGraphqlUrl } from '@snapwp/core';
```
justlevine, line 2: I hate this function and now its spreading to new parts of the codebase 😬

examples/nextjs/starter/codegen.ts
```diff
@@ -1,28 +1,13 @@
...
-			[ generateGraphqlUrl(
-				process.env.NEXT_PUBLIC_WORDPRESS_URL,
-				process.env.NEXT_PUBLIC_GRAPHQL_ENDPOINT
-			) ]: {
-				headers: {
-					Authorization: `${ process.env.INTROSPECTION_TOKEN }`,
-				},
-			},
-		},
-	],
```
justlevine, line 25: I don't like that we've removed the _entire_ schema.

There are many reasons that a project will need to change this, e.g. to change/add headers. (it's also one of the reasons I think the `generateGraphQLUrl` function should be removed).

justlevine (changes requested):
Need some clarification on the approach 🙇

packages/codegen-config/src/index.ts
```diff
@@ -1,13 +1,54 @@
...
+ */
+const baseConfig = (): CodegenConfig => {
+	return {
+		overwrite: true,
+		generates: {
+			'src/__generated/': {
+				preset: 'client',
+				plugins: [],
+			},
 		},
```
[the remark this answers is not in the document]
SH4LIN: Any common configurations from our end will be added directly to the `baseConfig()` function in the `codegen-config` package. Specific overriding configs can still be passed to `withCodegenConfig`, allowing them to override `baseConfig()`.

examples/nextjs/starter/codegen.ts
SH4LIN: There is still an option to override this by passing it as the config in `withCodegenConfig` function

packages/codegen-config/src/index.ts
SH4LIN: This function was used in two `codegen.ts` files (one of them being userland). We are now removing it from there and adding it to the `codegen-config` package, simplifying the userland `codegen.ts` and reducing its complexity.

examples/nextjs/starter/codegen.ts
SH4LIN: We are now moving the schema into our codegen package. If we continue using only the object, it incorrectly retrieves `process.env` values. To resolve this, I converted it into a function called `baseConfig()`, which first loads the process variables from individual `.env` files. After that, it calls the function, ensuring we get the correct `process.env` values.

For the `withCodegenConfig` function, we are following the same approach as we did for `withSnapWP`. We take the config from the userland `codegen.ts`, give it higher priority, and add our common configs, such as HOC. This ensures that all extensibility functionality remains intact.

This HOC approach allows us to modify internal functionality and update functions without requiring users to migrate between versions.  

For example, suppose we added the `generateGraphqlUrl` function in the userland `codegen.ts` and released the Next.js starter. Developers start using it, but later, we decide to change this function and introduce a new architecture. In that case, we would need to create a migration guide instructing users to adopt the new approach—despite them having no prior knowledge of the change. Additionally, we would have to follow a deprecation process for something that is purely an internal functionality.

packages/codegen-config/src/index.ts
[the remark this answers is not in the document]
justlevine: But users would need to override the entire object. With destructing you can do it at any level.

examples/nextjs/starter/codegen.ts
justlevine: Yes, but a user would need to override the _entire_ property...

packages/codegen-config/src/index.ts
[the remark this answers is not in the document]
SH4LIN: They can override any root-level property entirely, but they can't override a specific property within a root-level property.

examples/nextjs/starter/codegen.ts
SH4LIN: Can you give me an example of what you mean by an entire property?  

If you say they cannot override the `schema` property completely, they can. However, they won't be able to keep the `schema` property as our default and only change the `header`.

packages/codegen-config/src/index.ts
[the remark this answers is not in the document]
justlevine: Yes and that's a problem.

Look at https://github.com/AxeWP/axewp-frontend/blob/develop/graphql.config.ts and compare the properties there to our current codegens (plural).

Ideally a user is only overloading the config values they need, and not the entire parent tree.
Object destructuring empowers the user to overload the level they want (even if its messy). If you want to use an HOC then we need to optional-type it for the whole `CodegenConfig`, and **deep merge** it into our defaults.

SH4LIN:
@justlevine I have made changes to the code please re-review it, meanwhile I'll resolve merge conflict

packages/codegen-config/src/index.ts
```diff
@@ -1,13 +1,38 @@
 import type { CodegenConfig } from '@graphql-codegen/cli';
+import { generateGraphqlUrl } from '@snapwp/core';
```
justlevine, line 2: is `@snapwp/core` already a dependency of `codegen-config`? 

It shouldn't be.

justlevine (changes requested):
Some nitpicks, but then I think I need to defer to @ayushnirwal . I feel like we're adding tech-debt here without gaining anything.

Might help to see it with the merge-conflict resolved, since that's where the first pass of codegen types start getting added.

packages/codegen-config/src/index.ts
SH4LIN: @snapwp/core was dependency of `codegen-config` before this PR.

ayushnirwal (approved):
We are abstracting away the choice of a schema file or endpoint right now. This seems good enough solution to solve our problem of CI depending on a server.

The real solution is to manage this choice in the scaffold process or let the Users decide by exporting 2 configs one using schema file and other using an endpoint. That being said it is out of the scope of a pre-alpha release. Given the choice between a stability and ease of use. I would prefer to choose the later option now and move on.

PR #52, opened 2025-02-07
feat: Add better type definitions for block props

Swanand01:
## What

This PR adds type definitions for block prop types.

## Why

The exported block definitions can be used while overriding block definitions.

### Related Issue(s):

https://github.com/rtCamp/snapwp/issues/53

## How

The newly created prop types are exported from the component file. For now, the types are copied from the graphql generated types.

## Checklist


Swanand01:
@justlevine @ayushnirwal 
The block prop types generated by graphql have a lot of attributes that are optional.
Should the block renderer/component assume them to be optional?
Also, can we please have graphql return `undefined` for an optional attribute, instead of `undefined | null`

Swanand01:
https://github.com/rtCamp/snapwp/blob/94eccd492dbed367ec004b0706f8e6e47af295de/packages/blocks/src/blocks/tests/core-button.test.tsx#L158-L173

We cannot expect `expect( linkElement ).not.toHaveAttribute( 'href' ); ` to be true, since it is a required prop in `NextLink`, also, we are doing this: https://github.com/rtCamp/snapwp/blob/94eccd492dbed367ec004b0706f8e6e47af295de/packages/next/src/components/link.tsx#L66

justlevine:
> https://github.com/rtCamp/snapwp/blob/94eccd492dbed367ec004b0706f8e6e47af295de/packages/blocks/src/blocks/tests/core-button.test.tsx#L158-L173
> 
> We cannot expect `expect( linkElement ).not.toHaveAttribute( 'href' ); ` to be true, since it is a required prop in `NextLink`, also, we are doing this:
> 
> https://github.com/rtCamp/snapwp/blob/94eccd492dbed367ec004b0706f8e6e47af295de/packages/next/src/components/link.tsx#L66

I don't think I follow. Not all buttons are links, and not all `a` tags are `next/link`s either...

Seems the logic can be cleaned up in that component though.

packages/blocks/src/blocks/index.ts
```diff
@@ -63,4 +62,37 @@ const blocks: BlockDefinitions = {
 	default: Default,
 };
 
+export type BlockDefinitions = typeof blocks;
+
 export default blocks;
+
+export * from './core-audio';
```
justlevine, line 69: Really not a fan about how this just turned into a barrel file.

packages/blocks/src/blocks/index.ts
```diff
@@ -63,4 +62,37 @@ const blocks: BlockDefinitions = {
 	default: Default,
 };
 
+export type BlockDefinitions = typeof blocks;
```
justlevine, line 65: What does this mean for adding custom blocks that we haven't bundled by default?

packages/blocks/src/block-manager/index.tsx
```diff
@@ -1,23 +1,35 @@
-import {
-	type BlockData,
-	type BlockTreeNode,
-	type BlockDefinitions,
-} from '@snapwp/core';
 import defaultBlockDefinitions from '@/blocks';
 import flatListToHierarchical from '@/utils/flat-list-to-hierarchical';
 
+export interface BlockData {
+	type: keyof typeof defaultBlockDefinitions;
```
justlevine, line 5: The user is more likely to be creating components for blocks we don't bundle than for blocks we do.

packages/blocks/src/block-manager/index.tsx
```diff
@@ -1,23 +1,35 @@
...
+	cssClassNames?: Array< string | null > | null;
+	clientId?: string | null;
+	parentClientId?: string | null;
+	renderedHtml?: string | null;
+	attributes?: Record< any, any >;
```
justlevine, line 10: If the intent here is for attributes to get overloaded by the `{BlockComponent}` we should `unknown`, not any (and we know the key is always as string)

Swanand01:
@ayushnirwal @justlevine 
Do we care about prop type checking when the user is building the `blockDefinitions` object?
If no, we could define the type like this:
```ts
export type BlockDefinitions = {
    [ key: string ]: React.FC | null;
};
```

packages/blocks/src/blocks/index.ts
Swanand01: Updated the type for `BlockDefinitions`

packages/blocks/src/block-manager/index.tsx
Swanand01: Updated the type of `type` to `string`

packages/blocks/src/block-manager/index.tsx
```diff
@@ -1,17 +1,34 @@
...
+	type: string;
+	cssClassNames?: Array< string | null > | null;
+	clientId?: string | null;
+	parentClientId?: string | null;
+	renderedHtml?: string | null;
+	attributes?: T;
+}
+
+export type BlockTreeNode<
+	TBlockProps extends BlockData = BlockData< CoreVideoAttributes >,
```
ayushnirwal, line 20: `CoreVideoAttributes`?

packages/blocks/src/blocks/core-audio.tsx
```diff
@@ -1,9 +1,23 @@
...
+export interface CoreAudioAttributes extends Record< string, unknown > {
+	autoplay?: boolean;
+	caption?: string;
+	loop?: boolean;
+	preload?: string;
+	src?: string;
+	style?: string;
+}
+
+export interface CoreAudioProps extends Omit< BlockData, 'type' > {
```
ayushnirwal, line 18: Omit type before exporting `BlockData`

packages/blocks/src/blocks/core-audio.tsx
```diff
@@ -1,9 +1,23 @@
...
+import { BlockData } from '..';
+
+export interface CoreAudioAttributes extends Record< string, unknown > {
+	autoplay?: boolean;
+	caption?: string;
+	loop?: boolean;
+	preload?: string;
+	src?: string;
+	style?: string;
+}
```
ayushnirwal, line 16: Think carefully about what attributes should be optional. It does not make sense to render an Audio block without `src`

ayushnirwal (changes requested):

packages/blocks/src/blocks/core-audio.tsx
Swanand01: There is a check in the component:
https://github.com/rtCamp/snapwp/blob/110c3da7de0a6c9a489ae1568863f3f974b90cf4/packages/blocks/src/blocks/core-audio.tsx#L37-L39

I suppose returning null when no src is passed is the expected behaviour, since there is a test for the same:
https://github.com/rtCamp/snapwp/blob/110c3da7de0a6c9a489ae1568863f3f974b90cf4/packages/blocks/src/blocks/tests/core-audio.test.tsx#L66-L77

packages/blocks/src/blocks/core-audio.tsx
Swanand01: We can't do this, since `BlockData.type` is used here:
https://github.com/rtCamp/snapwp/blob/110c3da7de0a6c9a489ae1568863f3f974b90cf4/packages/blocks/src/block-manager/index.tsx#L68-L70

Should I create a separate type just for blocks?

Swanand01: Added a BlockProps type.

Swanand01:
> > This won't work anymore since `BlockDefinitions` and other block types have been moved to the `blocks` package. Moving these types back to `@core/props` is also not possible, since a dependency on the `blocks` package would be needed.
> 
> 
> 
> First sentence is the issue, and second sentence is the false premise.
> 
> 
> 
> The generic Block Definition types _should_ indeed be in `@snapwp/core`, so users can use any block library that's compatible with the API instead of our block package.
> 
> 
> 
> The trick here is to rely on interfaces/generics and then extend/implement them in blocks.
> 
> 
> 
> Only glanced at the diff, but I'm not seeing anything jump out at me that can't be decoupled into the above pattern. Ping me if y'all get stuck on a specific type.
> 
> 
> 
> 
> 
> 
> 
> 

We can't move the BlockTypeMap type to core, which is needed by BlockDefinitions type, since it imports block attributes.

packages/types/src/blocks/base.ts
```diff
@@ -0,0 +1,22 @@
...
+	type: string;
+	cssClassNames?: Array< string | null > | null;
+	clientId?: string | null;
+	parentClientId?: string | null;
+	renderedHtml?: string | null;
+	attributes?: T;
+}
+
+export type BlockProps<
+	T extends Record< string, unknown > = Record< string, unknown >,
```
SH4LIN, line 13: I think T should extend BlockData

packages/blocks/src/editor-blocks-renderer/index.tsx
```diff
@@ -24,15 +26,13 @@ export default function EditorBlocksRenderer( {
...
 			key: node.clientId,
 			...node,
+			renderer: undefined,
+			children: undefined,
 		};
 
-		// Removing renderer and children from props. Renderer should not be passed to the component.
-		delete props.renderer;
-		delete props.children;
-
```
ayushnirwal, line 35: This used to delete the keys too. Revert this change.

packages/query/package.json
```diff
@@ -45,6 +45,7 @@
 	"peerDependencies": {
 		"react": "^19.0.0",
 		"@apollo/client": "^3.11.4",
-		"graphql": "^16.9.0"
+		"graphql": "^16.9.0",
+		"@snapwp/types": "*"
```
ayushnirwal, line 49: I may be wrong but the types package should be a DevDependencies. The types should be in the build artifacts. Please test that theory out and move to DevDependencies if Im correct

@SH4LIN Opinions?

ayushnirwal (changes requested):
A Few minor changes required. Love to see the scope minimized back to the original scope of the issue ( Type Defs for Props and Attributes of Block Renderes ). Other types related to Block Manager class, Block overloading etc., should be handled later.

packages/blocks/src/block-manager/index.tsx
```diff
@@ -48,13 +45,22 @@ export default class BlockManager {
...
+
+		if ( customBlockDefinition === null ) {
+			// If explicitly set to null in custom definitions, use default renderer
+			node.renderer = BlockManager.blockDefinitions.default || Default;
+		} else if ( customBlockDefinition ) {
+			// If custom definition exists, use it
+			node.renderer = customBlockDefinition;
+		} else if ( defaultBlockDefinition ) {
+			// If no custom definition but default definition exists, use default
+			node.renderer = defaultBlockDefinition;
```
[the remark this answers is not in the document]
Swanand01: It doesn't hurt to leave this here, since this is the desired behaviour anyways.

packages/query/package.json
SH4LIN: This should be in `devDependencies`. The type used in the build artifact will be referenced by other TypeScript files. Since all TypeScript files are compiled during development and not executed at runtime, it should be a devDependency.

ayushnirwal (approved):

packages/core/src/props/index.ts
```diff
@@ -1,29 +1,3 @@
...
-export type BlockDefinitions = {
-	[ key: string ]: React.FC< BlockData >;
-};
-
-export type EditorBlocksRendererProps = {
-	editorBlocks?: BlockData[] | null;
-	blockDefinitions?: BlockDefinitions | null;
-};
-
 export type TemplateHeadProps = {
```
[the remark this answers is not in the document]
Swanand01: It is not in the scope of this PR. This PR only handles block type definitions.

SH4LIN (approved):
LGTM

PR #56, opened 2025-02-11
feat: add support for snapwp.config.ts

SH4LIN:
## What

This PR separates the config variables into two types:  

1. **EnvConfigs**  
2. **SnapWPConfigs**  

The key difference between these two is that **EnvConfigs** will contain all the configurations declared inside the `.env` file, which will be available during the Next.js build routine. In contrast, **SnapWPConfigs** will include configurations to override block definitions, queries, etc., and will only be available at runtime in Next.js.  

Additionally, this update introduces generic types to improve the type definitions for configurations.

## Why

With this change, we will support the TypeScript config file along with all other possible overrides provided through our config file.

### Related Issue(s):

- Part of https://github.com/rtCamp/headless/issues/403

## Checklist

### DIFF
--- a/examples/nextjs/starter/snapwp.config.mjs
+++ b/examples/nextjs/starter/snapwp.config.mjs
@@ -1,8 +1,0 @@
-// eslint-disable-next-line jsdoc/check-tag-names
-/** @type {import('@snapwp/core/config').SnapWPConfig} */
-const config = {
-	// Allow Proxy to WordPress assets (scripts, theme files, etc) to prevent CORS issues on localhost.
-	useCorsProxy: process.env.NODE_ENV === 'development',
-};
-
-export default config;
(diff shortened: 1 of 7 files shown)


justlevine (changes requested):
Code LGTM, but I need @ayushnirwal 's eye.

From a self-documentation perspective, I think we should change `SnapWPEnvConfig` to `SnapWPEnv`, that way it's more obvious that it parallels `SnapWPConfig` and (and isnt a child).  

(The internal `SnapWPConfigSchema` I don't care about since it's not exported, but I'd just call it `ConfigSchema`)

I split updating the docs into a subtask on https://github.com/rtCamp/headless/issues/403, so we can merge this and unblock the dependent tasks, while docs work happens in parallel.

ayushnirwal (approved):
LGTM

justlevine (approved):

PR #61, opened 2025-02-12
refactor: Update ESLint rulesets

Pathan-Amaankhan:
> [!Note]
> This pull request (PR) constitutes Part 2 of the migration process for the subsequent PRs: 
> - https://github.com/rtCamp/headless/pull/321
> - https://github.com/rtCamp/headless/pull/322
> - https://github.com/rtCamp/headless/pull/327

## What
- Update the ESLint rulesets and enforce stricter guidelines.

#### parser
- `@typescript-eslint/parser` parser migrated from `packages/eslint-config/index.js` as it is not required in public package because JavaScript files are shipped. (ref: https://github.com/rtCamp/snapwp/pull/60#issuecomment-2654116722)

#### ignorePatterns
- ignorePatterns are sorted ascending.

##### [@typescript-eslint/naming-convention](https://typescript-eslint.io/rules/naming-convention/)
- Enforcing naming conventions helps keep the codebase consistent, and reduces overhead when thinking about how to name a variable.

##### [@typescript-eslint/no-restricted-imports](https://typescript-eslint.io/rules/no-restricted-imports/)
- Restricts use of `classnames` & `lodash` imports in TypeScript files.

##### [dot-notation](https://eslint.org/docs/latest/rules/dot-notation)
- Enforces dot notation on [Objects](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object). Setting `allowKeywords` option to false prevents using reserved keywords.

##### [eslint-comments/require-description](https://mysticatea.github.io/eslint-plugin-eslint-comments/rules/require-description.html)
- Throws error if directive comments are used without description.

##### [import/default](https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/default.md)
- If a default import is requested, this rule will report if there is no default export in the imported module.

##### [import/named](https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/named.md)
- Verifies that all named imports are part of the set of named exports in the referenced module.

##### [jest/expect-expect](https://github.com/jest-community/eslint-plugin-jest/blob/main/docs/rules/expect-expect.md)
- Ensure that there is at least one expect call made in a test.

##### [jsdoc/no-types](https://github.com/gajus/eslint-plugin-jsdoc/blob/main/.README/rules/no-types.md)
- Allows adding type-definitions to `@param` & `@return`

[...]

PR #62, opened 2025-02-12
feat: add support for overloading `html-react-parser` options

BhumikP:
## What

This PR includes changes related allowing user to overload default html-react-parser option 

## Why

To let user have more control over options used  in html-react-parser

### Related Issue(s):

https://github.com/rtCamp/headless/issues/397

## Checklist


packages/core/src/props/index.ts
```diff
@@ -75,3 +75,15 @@ export type GlobalHeadProps = {
...
+	type: string;
+	name?: string;
+	children?: DOMNode[];
+	attribs?: Record< string, string >;
+	data?: string;
+};
+
+export type HTMLReactParserOptions = {
+	replace?: ( domNode: DOMNode ) => React.ReactElement | string | null | void;
+};
```
justlevine, line 89: I don't like the idea of manually maintaining this.

Not entirely sure how this works but can html-react-parser be a devDep so we can just recycle their type?

packages/core/src/config/snapwp-config-manager.ts
```diff
@@ -1,7 +1,7 @@
 'use snapWPConfig';
 import { isValidUrl, generateGraphqlUrl } from '@/utils';
 import { Logger } from '@/logger';
-import { BlockDefinitions } from '@/props';
+import { BlockDefinitions, HTMLReactParserOptions } from '@/props';
```
justlevine, line 4:
```suggestion
import type { BlockDefinitions, HTMLReactParserOptions } from '@/props';
```

packages/next/src/react-parser/index.tsx
```diff
@@ -97,5 +98,9 @@ const options: HTMLReactParserOptions = {
  * @return The rendered React components.
  */
 export default function Parse( { html }: { html: string } ) {
+	const { parserOptions } = getConfig();
+
+	const options = parserOptions ? parserOptions : defaultOptions;
+
 	return <>{ Parser( html, options ) }</>;
 }
```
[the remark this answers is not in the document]
BhumikP: do you mean move the default options to seperate  file ?

docs/overloading-wp-behavior.md
```diff
@@ -148,6 +148,133 @@ export default function Page() {
...
+					mediaDetails: {
+						width,
+						height,
+					},
+				};
+			const shouldFill =
+					! width &&
+					! height &&
+					undefined !== width &&
+					undefined !== height;
```
SH4LIN, line 209: I think For an example these things are not necessary, also in this example we are using some function from core API which we have marked as internal use.

docs/overloading-wp-behavior.md
```diff
@@ -148,6 +148,133 @@ export default function Page() {
...
+		}
+
+		return undefined;
+	},
+};
+```
+
+### 2. Use react-parser as fallback
+
+Make any default component null to use react-parser as fallback component
```
SH4LIN, line 238:
```suggestion
Make any default block component null to use react-parser as fallback component
```

SH4LIN (changes requested):

SH4LIN:
I tried adding a generic type, but we can't include it here unless we convert this config into a function that returns `HTMLParseOptions`, which would be overkill.

@justlevine @BhumikP @ayushnirwal

justlevine:
> unless we convert this config into a function that returns `HTMLParseOptions`, which would be overkill.

@SH4LIN why would this be overkill? It would definitely make it super easy to lazy-resolve custom options only on the routes that actually use the `Parse`r...

docs/overloading-wp-behavior.md
```diff
@@ -148,6 +148,133 @@ export default function Page() {
 
 ---
 
+## Overloading React Parser Options
+
+SnapWP allows you to pass custom options to [**react-parser**](../packages/next/src/react-parser/index.tsx) . This is useful when you want to overload default options of react-parser.
```
justlevine, line 153: There should be some mention and link to `html-react-parser`, "React Parser Options" isnt known terminology.

SH4LIN:
> @SH4LIN Why would this be overkill? 
Developers will need to return **parser options** from a function instead of passing them directly to the config. Wouldn't that hurt DX?

justlevine:
> > @SH4LIN Why would this be overkill?
> 
> Developers will need to return **parser options** from a function instead of passing them directly to the config. Wouldn't that hurt DX?

A bit, but if there's justification for it (e.g code splitting/lazyloading prerequisite) then there are ways to work around it.

E.g. could be addressed by calling it `parserOptionsFn()` or by bypassing this entirely and instead making the Parse component itself what's pluggable, instead of just the options.
(Those are just examples, not advocating for a specific solve until we have a valid use case)

SH4LIN:
@justlevine I created a POC for that. What I was referring to is that we need to create a function `() => {}` that returns our `HTMLParserOptions` object. This means wrapping our options inside a function, similar to how we did in `baseConfig()`. However, this solution itself doesn’t work. My conclusion is that we won’t be able to add a generic type to the options.

ayushnirwal:
> A bit, but if there's justification for it (e.g, code splitting/lazyloading prerequisite), there are ways to work around it.

Let's concentrate on adding ability to overload the parser for now. Patterns for splitting/lazy loading are required for block definitions too. We should make a new issue to collectively handle any overloading configurations in that.

packages/core/src/config/snapwp-config-manager.ts
```diff
@@ -2,7 +2,7 @@
 import { isValidUrl, generateGraphqlUrl } from '@/utils';
 import { Logger } from '@/logger';
 import type { BlockDefinitions } from '@/props';
-
+import type { HTMLReactParserOptions } from 'html-react-parser';
```
ayushnirwal, line 5: @SH4LIN / @justlevine How about a prettier rule to add newline after import statements?

ayushnirwal (approved):
LGTM

packages/core/src/config/snapwp-config-manager.ts
justlevine: Make a 1-off PR against https://github.com/rtCamp/headless/issues/250

Though this might be covered by ESlint's (non-wp) Dependency groupings, ref https://github.com/rtCamp/snapwp/pull/61/files/24fd3d30122ada52b7836cdf5f13d1ae589843f2#r1958860096

docs/overloading-wp-behavior.md
justlevine: @BhumikP bumping

docs/overloading-wp-behavior.md
justlevine: @BhumikP bumping

docs/overloading-wp-behavior.md
justlevine: @BhumikP bumping

justlevine:
@BhumikP this looks good 🚀

Can you [add a changeset](https://github.com/rtCamp/snapwp/blob/develop/DEVELOPMENT.md#changesets)?

packages/core/package.json
```diff
@@ -49,6 +49,7 @@
 		"@types/react": "*",
 		"resolve-tspaths": "*",
 		"tsc-watch": "*",
-		"typescript": "*"
+		"typescript": "*",
+		"html-react-parser": "^5.1.12"
```
justlevine, line 53: Not entirely sure how this works, do we want to manage this separately from next, or could we use a "*" since we only care about the type? 

cc @ayushnirwal

docs/overloading-wp-behavior.md
```diff
@@ -149,6 +149,115 @@ export default function Page() {
 
 ---
 
+## Overloading [html-react-parser](https://www.npmjs.com/package/html-react-parser) Options
```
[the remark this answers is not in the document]
BhumikP: Done

SH4LIN (dismissed):

docs/overloading-wp-behavior.md
```diff
@@ -4,6 +4,139 @@ SnapWP takes an additive approach to headless WordPress development, where WordP
...
+ * Ref : https://github.com/remarkablemark/html-react-parser/issues/221#issuecomment-784073240
+ * @param domNode The DOM element
+ * @return parser options
+ */
+const isElement = ( domNode: DOMNode ): domNode is Element => {
+	const isTag = domNode.type === 'tag';
+	const hasAttributes = ( domNode as Element ).attribs !== undefined;
+
+	return isTag && hasAttributes;
+};
```
justlevine, line 60: Should we be doing this in _our_ `options`? If so, makes sense to export this function and update the docs (in a follow up PR)

justlevine, line 60: Should we be doing this in _our_ options?

(If so we should implement this ourselves, export the function for userland, and update the example)

BhumikP: I don't think we should add this in our implementation. It should be completely dependent on user's choice how they want to solve it by checking Ref. link.

PR #63, opened 2025-02-13
build: update base tsconfig rules

BhumikP:
## What

This PR includes strict type checking by adding some rules in tsconfig.base.json 

```
 "noFallthroughCasesInSwitch": true, // pevents execution from falling through to the next case in a switch without an explicit break
 "noUnusedParameters": true,   // Throws an error if a function parameter is declared but never used.
 "noUnusedLocals": true, // Throws an error if a local variable is declared but never used.
 "strictFunctionTypes": true, //   Enforces stricter type checking for function parameters and return types to prevent unsafe assignments.
```
### Related Issue(s):

https://github.com/rtCamp/headless/issues/317

## Checklist

-   [] My code has detailed inline documentation.


packages/blocks/src/blocks/core-image.tsx
```diff
@@ -55,7 +55,7 @@ interface MediaSize {
 	name: string;
 	sourceUrl: string;
 }
-
+//@ts-ignore
```
justlevine, line 58: Anywhere we `@ts-ignore` we should leave a reason  e.g. 
```suggestion
// @ts-ignore: Unused until lightbox support is fixed.
```

justlevine (dismissed):
A single nitpick

justlevine (approved):
Only change in https://github.com/rtCamp/snapwp/pull/63/commits/c1a7f1fb9e09bfc44aecbca1f9e3f632bf6f7c4c was formatting the `compilerOptions` (per the TS spec), no changes to properties.

PR #64, opened 2025-02-13
chore: implement additional tsconfig properties

justlevine:
## What

This PR makes the following changes to `tsconfig.base.json`. Reasons are included alongside the change:

- `noImplicitOverride: true` - We're using classes, and we may want to extend them in future packages
- `declaration: true` - We're a library. `d.ts` files help end user IDEs
- `sourceMap: true` -  We're a library, sourcemaps help end users debug
- `isolatedModules:  true` - limits some ts edge cases.
- `moduleDetection: "force"` - Prevents unexpected behavior leaking from autodetection.
- `resolveJsonModule: true` - so we can import .json
- `module|target: es2022` - the modern "stable" version.

## Why

These are all noncontroversial, incremental improvements that don't require any existing code changes to satisfy.

Individual property explanations are here https://www.typescriptlang.org/tsconfig

### Related Issue(s):

Part of https://github.com/rtCamp/headless/issues/317

## Additional Info

Additionally set `noUncheckedIndexedAccess` and `verbatimModuleSyntax` to false, but they'll be handled in a separate PR since they require file changes.

## Checklist

### DIFF
--- a/config/tsconfig.base.json
+++ b/config/tsconfig.base.json
@@ -28,7 +28,7 @@
 		"esModuleInterop": true,
 		"forceConsistentCasingInFileNames": true,
 		"isolatedModules": true,
-		"verbatimModuleSyntax": false,  // @todo Enable this.
+		"verbatimModuleSyntax": false, // @todo Enable this.
 
 		// Language and Environment
 		"jsx": "react",
--- a/config/tsconfig.base.json
+++ b/config/tsconfig.base.json
@@ -7,17 +7,20 @@
 		"allowSyntheticDefaultImports": true,
 		"allowUnreachableCode": false,
 		"allowUnusedLabels": false,
+		"exactOptionalPropertyTypes": false, // @todo Enable this.
 		"noFallthroughCasesInSwitch": true,
 		"noImplicitOverride": true,
 		"noImplicitReturns": true,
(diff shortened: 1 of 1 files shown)


config/tsconfig.base.json
```diff
@@ -18,16 +20,21 @@
...
 		"forceConsistentCasingInFileNames": true,
+		"isolatedModules": true,
+		"verbatimModuleSyntax": false, // @todo Enable this.
 
 		// Language and Environment
 		"jsx": "react",
 		"lib": [ "ES2022", "DOM" ],
-		"target": "ES2018",
+		"moduleDetection": "force",
+		"target": "ES2018", // @todo bump to ES2022
```
justlevine, line 37: Is there a reason _not_ to bump this before 0.1.0?

ayushnirwal: Nope. Bump it.

ayushnirwal (approved):

justlevine:
https://github.com/rtCamp/snapwp/pull/64/commits/5dbe955e2208bece6fa372b8da4599351e5c351f adds

- `resolveJsonModule: true`
- bumps `module` and `target` to `ES2022`
- stubs `exactOptionalPropertyTypes` and `noPropertyAccessFromIndexSignature` (enabling would require file remediation)

PR #74, opened 2025-02-18
fix: links breaking when there is no trailing slash in `NEXT_PUBLIC_URL`

Swanand01:
## What

This PR fixes the bug where links would break when there is no trailing slash in `NEXT_PUBLIC_URL` or `NEXT_PUBLIC_WORDPRESS_URL`

## Why

Links would not work.

## How

`normalizeConfig` is updated to remove trailing slashes from `homeUrl` and `nextUrl`.

## Checklist

### DIFF
--- a/packages/core/src/config/snapwp-config-manager.ts
+++ b/packages/core/src/config/snapwp-config-manager.ts
@@ -220,6 +220,7 @@ class SnapWPConfigManager {
 				if ( cfg[ key ] === undefined ) {
 					delete cfg[ key ];
 				} else if (
+					// This should probably be moved into the schema as a sanitize callback.
 					( key === 'homeUrl' || key === 'nextUrl' ) &&
 					typeof cfg[ key ] === 'string'
 				) {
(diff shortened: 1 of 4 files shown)


packages/core/src/config/snapwp-config-manager.ts
```diff
@@ -220,6 +220,7 @@ class SnapWPConfigManager {
 				if ( cfg[ key ] === undefined ) {
 					delete cfg[ key ];
 				} else if (
+					// This should probably be moved into the schema as a sanitize callback.
```
justlevine, line 223: It's the keyword that allows these to be searchable i.e. actionable.

(if we still havnt fixed the eslint syntax warning then `// TODO:` is a good replacement for `// @todo`

```suggestion
					// @todo This should probably be moved into the schema as a sanitize callback.
```

justlevine:
Also @Swanand01, can you [add a changeset](https://github.com/rtCamp/snapwp/blob/develop/DEVELOPMENT.md#changesets)?

packages/core/src/config/snapwp-config-manager.ts
Swanand01: Oh no, my bad. Forgot to add the keyword.

ayushnirwal (changes requested):
LGTM.

Just add a test case for this in config manager unit tests.

justlevine (approved):

ayushnirwal (approved):

PR #76, opened 2025-02-18
chore: remove example package-lock file from VCS

justlevine:
## What

Removes `examples/**/package-lock.json` from the GH repository

## Why

It makes no sense to VCS this when we want the user to install whatever the latest versions are via `npx snapwp`.

Maintaining the lockfile manually (especially after bumping versions with changesets) is not sustainable.

## Checklist


Pathan-Amaankhan (approved):
LGTM

PR #82, opened 2025-02-20
Docs: add manual installation steps

ayushnirwal:
## What

Adds documentation for manually using `@snapwp` libraries in a next js app.

## Checklist

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -74,75 +74,83 @@ To create a new headless WordPress app using SnapWP, follow these steps:
 
 ### Manual Installation
 
-1. Install npm packages in your Next App
+1. Install the required packages in your Next.js project:
 
-```bash
-npm install @snapwp/blocks @snapwp/core @snapwp/next @snapwp/query
-```
+    ```bash
+    npm install --save @snapwp/blocks @snapwp/core @snapwp/next @snapwp/query
+    ```
 
-2. Create an `.env` file in the project root. The contents for the .env file can be copied from `Dashboard > WPGraphQL > Settings > SnapWP Helper` from the WP server.
+2. Create an `.env` file in the project root. The contents for the .env file can be copied from `Dashboard > WPGraphQL > Settings > SnapWP Helper` from your [WordPress backend](#backend-setup).
 
 3. Create `snapwp.config.ts` in the project root with the following content.
 
-```typescript
-import type { SnapWPConfig } from '@snapwp/core/config';
+    ```typescript
+    // snapwp.config.ts
+    import type { SnapWPConfig } from '@snapwp/core/config';
 
-const config: SnapWPConfig = {};
+    const config: SnapWPConfig = {};
 
-export default config;
-```
+    export default config;
+    ```
 
-4. Update the nextjs config. Rename the file to `next.config.mjs` so that top-level await can be used.
+4. Update your [NextJS Config file](https://nextjs.org/docs/api-reference/next.config.js/introduction).
 
-```javascript
-import withSnapWP from '@snapwp/next/withSnapWP';
+    **Important**: Make sure to use the `*.mjs` filetype for your config (i.e `next.config.mjs` ) so that top-level `await` can be used.
 
-export default await withSnapWP( {} );
-```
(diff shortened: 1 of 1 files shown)


docs/getting-started.md
```diff
@@ -74,7 +74,75 @@ To create a new headless WordPress app using SnapWP, follow these steps:
 
 ### Manual Installation
 
-@todo
+1. Install npm packages in your Next App
+
+```bash
+npm i @snapwp/blocks @snapwp/core @snapwp/next @snapwp/query
```
justlevine, line 80: In user-facing docs we should avoid aliases.

Also, do we want a save flag so these are stored in their package.json?

```suggestion
npm install @snapwp/blocks @snapwp/core @snapwp/next @snapwp/query
```

docs/getting-started.md
```diff
@@ -74,7 +74,75 @@ To create a new headless WordPress app using SnapWP, follow these steps:
...
+```
+
+2. Copy the .env contents from `Dashboard > WPGraphQL > Settings > SnapWP Helper` from the WP server and create an `.env` file in the project root.
```
justlevine, line 83: Reverse the order: first you create an env file, and then you copy-paste the contents into it.

docs/getting-started.md
```diff
@@ -74,7 +74,75 @@ To create a new headless WordPress app using SnapWP, follow these steps:
...
+export default function Layout( { children }: { children: React.ReactNode } ) {
+	return (
+		<RootLayout>
+			<>{ children }</>
+		</RootLayout>
+	);
+}
+```
+
+6. Update the nextjs config. Rename the file to `next.config.mjs` so that top-level await can be used.
```
justlevine, line 127: This feels to me like it should be after step 3, to keep the config steps together

docs/getting-started.md
```diff
@@ -74,7 +74,75 @@ To create a new headless WordPress app using SnapWP, follow these steps:
...
+
+```javascript
+import withSnapWP from '@snapwp/next/withSnapWP';
+
+export default await withSnapWP( {} );
+```
+
+7. Check if you are using webpack for dev builds. Currently we don't support turbopack for dev builds. Remove the flag `--turbopack` to opt-out of using turbopack.
```
justlevine, line 135: I think we format this as a `Note:` not a numbered step, 
along the lines of currently we "only support webpack" (vs dont support turbopack), if you're currently using turbopack, make sure to remove the `--turbopack` flag in your `package.json`

justlevine (changes requested):

justlevine (approved):
@ayushnirwal I cleaned up some of the formatting, and added a note about conflicting routes per @ashutoshgautams 's feedback.

@ashutoshgautams please approve or leave comments where you feel the steps are still potentially confusing 🙇

docs/getting-started.md
```diff
@@ -74,7 +74,83 @@ To create a new headless WordPress app using SnapWP, follow these steps:
...
 ### Manual Installation
 
-@todo
+1. Install the required packages in your Next.js project:
+
+    ```bash
+    npm install --save @snapwp/blocks @snapwp/core @snapwp/next @snapwp/query
+    ```
+
+2. Create an `.env` file in the project root. The contents for the .env file can be copied from `Dashboard > WPGraphQL > Settings > SnapWP Helper` from your [WordPress backend](#backend-setup).
```
ashutoshgautams, line 83: I would recommend 
```md
Dashboard > GraphQL > SnapWP
```
image

I have attached a screenshot for flow reference. What do you think ?
It is a minor thing, but `Dashboard > WPGraphQL > Settings > SnapWP Helper` this makes me look twice even though I am sure I am doing the right thing.

ashutoshgautams (approved):
Overall, LGTM 🚀 

I have recommended a change for clarity but since it is minor/ my personal opinion,
we can either reject or accept the request and are good to merge.

docs/getting-started.md
justlevine: This feels like overkill, it isn't the focus of the section, and the previous screenshot already shows the menu options in context.

PR #83, opened 2025-02-20
feat: allow overloading of query from snapwp configs

SH4LIN:
## What

- This PR will allow developers to pass their generated query document to the config in order to execute their own query instead of our own.

## Caution
- This will allow developers to pass their own query, but we can't enforce any restrictions on its format. There is a high chance that if a developer does not structure the query as expected, the frontend will break.

### Related Issue(s):

- https://github.com/rtCamp/headless/issues/402

## How

- Created a config property to which developers can pass their query document which will be used to execute the graphql query

## Testing Instructions

- Create a new query inside the /examples/nextjs/starter/src
- Execute npm run codegen
- Pass the generated document from the config.

## Checklist


packages/types/package.json
```diff
@@ -38,6 +38,7 @@
 		"@types/react": "*",
 		"resolve-tspaths": "*",
 		"tsc-watch": "*",
-		"typescript": "*"
+		"typescript": "*",
+		"@apollo/client": "^3.11.4"
```
justlevine, line 42: Should `@apollo/client` be a devDep or an _optional_ peer dep?

packages/types/src/queries/global-styles.ts
```diff
@@ -0,0 +1,4 @@
+import type { TypedDocumentNode } from '@apollo/client';
+
+// Until we figure out generic types for GraphQL queries, we'll use `any` for now.
```
justlevine, line 3: Why `any` vs `unknown`?

packages/types/package.json
ayushnirwal: This is a `devDep` only required when building.

ayushnirwal:
Is this on hold?

justlevine:
> Is this on hold?

Not anymore (was just bumped from v0.1.0), but we did want to revisit it in context of more general changes we need to make to the QueryEngine

PR #60, opened 2025-02-12
refactor: Update eslint import-resolver for TS

Pathan-Amaankhan:
> [!Note]
> This pull request (PR) constitutes Part 1 of the migration process for the subsequent PRs: 
> - https://github.com/rtCamp/headless/pull/321
> - https://github.com/rtCamp/headless/pull/322
> - https://github.com/rtCamp/headless/pull/327

## What
- This PR migrates import-resolver from eslint-config package to `config/` directory.

### Related Issue(s):
- https://github.com/rtCamp/headless/issues/250

## Testing Instructions
- Run `npm install && npm run lint`.

## Checklist

-   [x] I have read the [Contribution Guidelines](https://github.com/rtCamp/snapwp/blob/develop/.github/CONTRIBUTING.md).
-   [x] My code is tested to the best of my abilities.
-   [x] My code passes all lints (ESLint, tsc, prettier etc.).
-   [x] My code has detailed inline documentation.
-   [ ] I have added unit tests to verify the code works as intended.
-   [ ] I have updated the project documentation accordingly.

PR #84, opened 2025-02-20
chore: cleanup ESLint rulesets

Pathan-Amaankhan:
## What
- This PR reformats eslint rulesets i.e. migrate configurations to their respective places & adds following plugins + rulesets:

#### parser
- `@typescript-eslint/parser` parser migrated from `packages/eslint-config/index.js` as it is not required in public package because JavaScript files are shipped. (ref: https://github.com/rtCamp/snapwp/pull/60#issuecomment-2654116722)

#### ignorePatterns
- ignorePatterns are sorted ascending.

##### [@typescript-eslint/no-restricted-imports](https://typescript-eslint.io/rules/no-restricted-imports/)
- Restricts use of `classnames` & `lodash` imports in TypeScript files.

##### [import/default](https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/default.md)
- If a default import is requested, this rule will report if there is no default export in the imported module.

##### [import/named](https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/named.md)
- Verifies that all named imports are part of the set of named exports in the referenced module.

##### [jest/expect-expect](https://github.com/jest-community/eslint-plugin-jest/blob/main/docs/rules/expect-expect.md)
- There can be test cases that might not have expect call made.

##### [jsdoc/no-types](https://github.com/gajus/eslint-plugin-jsdoc/blob/main/.README/rules/no-types.md)
- Allows adding type-definitions to `@param` & `@return`

##### [no-restricted-imports](https://eslint.org/docs/latest/rules/no-restricted-imports)
- Restricts use of `classnames` & `lodash` imports in JavaScript files.

##### [no-restricted-syntax](https://eslint.org/docs/latest/rules/no-restricted-syntax)
- Disallows specified syntax (Currently no syntax is restricted but as the project grows certain restrictions will be added as required in future versions)

#### overrides
- Overrides are migrated from `packages/eslint-config/index.js`, no additional override is added.

#### plugins

The following new plugins have been added that will be used to make strict rules in upcoming PRs:

- [@typescript-eslint](https://www.npmjs.com/package/@typescript-eslint/eslint-plugin)
- [jest](https://www.npmjs.com/package/eslint-plugin-jest)
- [@stylistic/jsx](https://www.npmjs.com/package/@stylistic/eslint-plugin-jsx)

## Related Issue(s):

## Testing Instructions

## Additional Info
- This PR is a breakdown of https://github.com/rtCamp/snapwp/pull/61. As we are just migrating configs from one file to another & adding rulesets that the code is current following, we won't be getting any eslint errors in this PR.
- We will raise ruleset (causing eslint errors) wise PRs, along with their fixes.

## Checklist

### DIFF
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -5,6 +5,7 @@ module.exports = {
 	},
 	extends: '@snapwp/eslint-config',
 	parser: '@typescript-eslint/parser',
+	plugins: [ '@typescript-eslint', 'jest', '@stylistic/jsx' ],
 	ignorePatterns: [
 		'**/__generated/',
 		'**/.eslintrc.cjs',
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -4,6 +4,7 @@ module.exports = {
 		node: true,
 	},
 	extends: '@snapwp/eslint-config',
+	parser: '@typescript-eslint/parser',
 	ignorePatterns: [
 		'**/__generated/',
 		'**/.eslintrc.cjs',
@@ -22,6 +23,43 @@ module.exports = {
 	},
 	settings: {
 		'import/resolver': require.resolve( './config/import-resolver.cjs' ),
+	},
+	rules: {
+		// Restrict usage of "bad" libraries.
+		'@typescript-eslint/no-restricted-imports': [
+			'error',
+			'classnames',
+			'lodash',
+		],
+		'no-restricted-imports': [
+			'error',
+			{
+				paths: [
+					{
+						name: 'classnames',
+						message:
+							"Please use `clsx` instead. It's a lighter and faster drop-in replacement for `classnames`.",
+					},
+					{
+						name: 'lodash',
+						message: 'Please use native functionality instead.',
+					},
+				],
+			},
+		],
+
+		// Restricted syntax should error, not warn.
+		'no-restricted-syntax': [ 'error' ],
+
+		// Import rules.
+		'import/default': 'error',
+		'import/named': 'error',
+
+		// Jest rules.
+		'jest/expect-expect': 'off',
+
+		// Turn of JSdoc types and use TypeScript types instead.
+		'jsdoc/no-types': [ 'off' ],
 	},
 	overrides: [
 		{
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -5,7 +5,7 @@ module.exports = {
 	},
 	extends: '@snapwp/eslint-config',
 	parser: '@typescript-eslint/parser',
-	plugins: [ '@stylistic/jsx', '@typescript-eslint', 'jest' ],
+	plugins: [ '@typescript-eslint', 'jest' ],
 	ignorePatterns: [
 		'**/__generated/',
 		'**/.eslintrc.cjs',
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -5,7 +5,7 @@ module.exports = {
 	},
 	extends: '@snapwp/eslint-config',
 	parser: '@typescript-eslint/parser',
-	plugins: [ '@typescript-eslint', 'jest', '@stylistic/jsx' ],
+	plugins: [ '@stylistic/jsx', '@typescript-eslint', 'jest' ],
 	ignorePatterns: [
 		'**/__generated/',
 		'**/.eslintrc.cjs',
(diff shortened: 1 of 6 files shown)


justlevine (commented):
@Pathan-Amaankhan https://github.com/rtCamp/snapwp/pull/84/commits/bf4ec648ee120eb9897fe09d4b49806e5a391e23 is the remaining diff after cherrypicking into #89 

Remaining questions being:

- Why are we adding these plugins at all.
- When adding these plugins, should we be adding them to _our_ rulesets or forcing them on the entire ecosystem via the package
- If they're being added in the package, shouldn't they be package dependencies (vs root-level ones)?

Pathan-Amaankhan:
> Why are we adding these plugins at all?
- Rules from the added plugins will be used in the upcoming PRs.


> When adding these plugins, should we be adding them to our rulesets or forcing them on the entire ecosystem via the package?
- This depends on the rulesets we choose to add to `@snapwp/eslint-config` package.
- If we choose to add any plugin-specific ruleset in `@snapwp/eslint-config` package, we would require the corresponding plugin to be added in our package.
- Fow now, as only `n/no-process-env` rule is added (that was also previously available), the only plugin on which our `@snapwp/eslint-config` depends is `n`.

> If they're being added in the package, shouldn't they be package dependencies (vs root-level ones)?
- For now, as we are not clear whether our plugins will make their way to `@snapwp/eslint-config` package, have added their package dependencies to the internal `package.json`.

All of the above questions will have a more precise answers, once we finalize what all things need to be migrated to our `@snapwp/eslint-config` package.

---

- Also one more thing, I am also not sure if we should keep our plugins in packages or migrate them to the internal eslint file (to prevent additional plugin changeset from being triggered) for now & migrate them once rulesets (that need to be added in `@snapwp/eslint-config`) are final. Please let me know your opinion on this.

justlevine:
> > Why are we adding these plugins at all?
> 
> * Rules from the added plugins will be used in the upcoming PRs.

I think we wait to add dependencies until we _use_ dependencies. Responding to the rest below, but I think this is enough to close this PR in favor of the "upcoming" one(s).

---

> 
> > When adding these plugins, should we be adding them to our rulesets or forcing them on the entire ecosystem via the package?
> 
> * This depends on the rulesets we choose to add to `@snapwp/eslint-config` package.
> [...]
> All of the above questions will have a more precise answers, once we finalize what all things need to be migrated to our `@snapwp/eslint-config` package.

Currently you added them in _our_ package.json, but marked them as plugins inside the `@snapwp/eslint-config` package, where they're _not_ installed. 


> * Also one more thing, I am also not sure if we should keep our plugins in packages or migrate them to the internal eslint file (to prevent additional plugin changeset from being triggered) for now & migrate them once rulesets (that need to be added in `@snapwp/eslint-config`) are final. Please let me know your opinion on this.

Didn't understand this question.

package.json
```diff
@@ -85,7 +85,11 @@
 		"resolve-tspaths": "^0.8.23",
 		"ts-jest": "^29.2.5",
 		"tsc-watch": "^6.3.0",
-		"typescript": ">=4.3.5 <5.4.0"
+		"typescript": ">=4.3.5 <5.4.0",
+		"@typescript-eslint/eslint-plugin": "^6.4.1",
+		"@typescript-eslint/parser": "^6.0.0",
+		"@stylistic/eslint-plugin-jsx": "^3.1.0",
+		"eslint-plugin-eslint-comments": "^3.2.0"
```
justlevine, line 92: Did you mean to leak this?

.eslintrc.cjs
```diff
@@ -5,6 +5,7 @@ module.exports = {
 	},
 	extends: '@snapwp/eslint-config',
 	parser: '@typescript-eslint/parser',
+	plugins: [ '@stylistic/jsx', '@typescript-eslint', 'jest' ],
```
justlevine, line 8: Drop stylistic from this PR in favor of #93

justlevine (changes requested):

package.json
Pathan-Amaankhan: Have resolved the leaking issue in following PRs and will be doing the same for the rest:
- https://github.com/rtCamp/snapwp/pull/98
- https://github.com/rtCamp/snapwp/pull/99

.eslintrc.cjs
Pathan-Amaankhan: Removed in https://github.com/rtCamp/snapwp/pull/84/commits/0a699d01d7d2eca75f72820a8daa82e424140dc0

justlevine (approved):

justlevine (approved):

PR #85, opened 2025-02-20
fix: address CodeQL smells in GitHub workflows

L0RD-ZER0:
## What
This PR Addresses CodeQL findings in the github workflows
In addition, fixes issue of diverging histories when publishing changes to `main` branch.

## Why
It is essential to address the issue of code quality in the ci to ensure a smoother development process. 

## Additional Info
Please refer to [security section](https://github.com/rtCamp/snapwp/security) for more info on existing CodeQL violations.

## Checklist


mrrobot47 (approved):

.github/workflows/generate-schema.yml
```diff
@@ -45,15 +48,15 @@ jobs:
                   node-version-file: '.nvmrc'
 
             - name: Setup PHP
-              uses: shivammathur/setup-php@v2
```
justlevine, line 48: how serious are these `@{version}` violations?

I don't want our CI turning into tech debt, and a `vX` makes it easy to see at a glance what needs bumping when the time comes.

I defer to @mrrobot47

L0RD-ZER0: The thought here was to pin the hash, and then in the future add dependot to run only for GitHub actions and get the updation work done.

c.c. @mrrobot47

justlevine: (A use case for dependabot I don't absolutely hate 🙌🍻, still deferring to Riddhesh)

mrrobot47: These violations are medium security IMO. Better to have pinned commit hash.
I had discussed with @L0RD-ZER0 to setup dependabot for workflow updates post this PR. _Although, ideally we should check the code of action for bump as well when dependabot updates the action. But generally we would be fine with the dependabot updates._

mrrobot47: https://docs.github.com/en/actions/security-for-github-actions/security-guides/security-hardening-for-github-actions#using-third-party-actions

mrrobot47: We can skip pinning rtCamp's action from the workflow to commit hash as they are first party GitHub actions for us.

mrrobot47: Also, @L0RD-ZER0 when you are pinning them, better to add a comment with the version like
```yml
uses: shivammathur/setup-php@9e72090525849c5e82e596468b86eb55e9cc5401 # Pinned @v2.23.0 
```

justlevine (approved):

PR #86, opened 2025-02-21
refactor!: move ESLint overrides from package to local ruleset

justlevine:
## What

This PR removes the Eslint config `overrides` from `@snapwp/eslint-config` and relocates them into our own package.

## Why

We want our package to implement smart and measured rules that make sense contextually, but these leaked into the initial release. 

They may even be restored as a ruleset.

### Related Issue(s):

- Cherrypicked from https://github.com/rtCamp/snapwp/pull/84
- Part of https://github.com/rtCamp/headless/issues/250

## Additional Info

This is a breaking change as those relying on those previous overrides will now need to override them themselves.

## Checklist

### DIFF
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -4,6 +4,7 @@ module.exports = {
 		node: true,
 	},
 	extends: '@snapwp/eslint-config',
+	parser: '@typescript-eslint/parser',
 	ignorePatterns: [
 		'**/__generated/',
 		'**/.eslintrc.cjs',
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -5,16 +5,17 @@ module.exports = {
 	},
 	extends: '@snapwp/eslint-config',
 	ignorePatterns: [
-		'**/node_modules/**',
-		'**/dist/**',
+		'**/__generated/',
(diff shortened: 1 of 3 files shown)


.eslintrc.cjs
```diff
@@ -23,7 +24,32 @@ module.exports = {
...
+						require: {
+							ArrowFunctionExpression: true,
+							ClassDeclaration: true,
+							ClassExpression: true,
+							FunctionExpression: true,
+							MethodDefinition: true,
+						},
+					},
+				],
+				'import/default': [ 'off' ],
```
justlevine, line 49: @Pathan-Amaankhan can you clarify:
- why do _we_ want this `off``?
- why do people who use `@snapwp/eslint-config` _not_ want this `off`? 
🙇

Pathan-Amaankhan: Hi @justlevine,
As the original PR is just a migration/re-formatting PR & adds non-error throwing rulesets, this change was to be removed in subsequent override/ruleset PR (& to make it same as had in [original PR](https://github.com/rtCamp/snapwp/pull/61/files)), so that we can also tackle the code changes in the same PR (if any need to be done).

PR #90, opened 2025-02-21
refactor: implement TS exactOptionalPropertyType rule

BhumikP:
## What

Implemention of ```"exactOptionalPropertyTypes": true``` rule

## Why

 stricter rules around how ts handles properties on type or interfaces which have a ? prefix.

### Related Issue(s):

https://github.com/rtCamp/headless/issues/317

## Additional Info

```
interface UserDefaults {
  colorThemeOverride?: "dark" | "light";
}
declare function getUserSettings(): UserDefaults;
// ---cut---
const settings = getUserSettings();
settings.colorThemeOverride = "dark";
settings.colorThemeOverride = "light";

// But not:
settings.colorThemeOverride = undefined;
//type 'undefined' is not assignable to type '"dark" | "light"' with 'exactOptionalPropertyTypes: true'
```
Without this flag enabled, there are three values which you can set colorThemeOverride to be: “dark”, “light” and undefined .

## Checklist


packages/blocks/src/blocks/core-audio.tsx
```diff
@@ -25,7 +25,7 @@ const CoreAudio: CoreAudioType = ( {
 		return null;
 	}
 
-	const styleObject = getStylesFromAttributes( { style } );
+	const styleObject = style ? getStylesFromAttributes( { style } ) : {};
```
justlevine, line 28: These empty objects (vs `undefined` or null) concern me. We don't want <tag style="" />, if there's no style object, we want `<tag />` _without_ the empty `style` attribute.

Same logic applies for any other "falsey object" and any other optional prop.

packages/core/src/config/snapwp-config-manager.ts
```diff
@@ -71,22 +68,40 @@ const defaultConfig: Partial< SnapWPEnv & SnapWPConfig > = {
...
+/**
+ * Default configuration.
+ */
+const defaultConfig: Partial< SnapWPEnv & SnapWPConfig > = {
+	graphqlEndpoint: DEFAULTS.graphqlEndpoint,
+	uploadsDirectory: DEFAULTS.uploadsDirectory,
+	restUrlPrefix: DEFAULTS.restUrlPrefix,
+	corsProxyPrefix: DEFAULTS.corsProxyPrefix,
+	useCorsProxy: DEFAULTS.useCorsProxy,
+};
```
justlevine, line 80: How is this (and the duplicate `DEFAULTS` ) either helpful or necessary?

Seems like we're just trying to appease a linter, which is always a bad sign.

packages/next/src/components/link.tsx
```diff
@@ -12,6 +12,7 @@ interface LinkInterface {
 	style?: CSSProperties | undefined;
 	className?: string | undefined;
 }
+interface NavLinkProps extends LinkProps, HTMLAttributes< HTMLAnchorElement > {}
```
justlevine, line 15: Whitespace is free

```suggestion
}

interface NavLinkProps extends LinkProps, HTMLAttributes< HTMLAnchorElement > {}
```

packages/next/src/components/link.tsx
```diff
@@ -30,9 +31,7 @@ export default function Link( {
 	className,
 	children,
 	...props
-}: PropsWithChildren<
-	LinkInterface & ( LinkProps | AnchorHTMLAttributes< HTMLAnchorElement > )
-> ) {
+}: PropsWithChildren< LinkInterface & NavLinkProps > ) {
```
justlevine, line 34: Previously the type was a union **either** a `next/link` _or_ an `a` tag, which I believe is what we want... why would we instead want to clobber the types together?

packages/next/src/react-parser/options.tsx
```diff
@@ -70,13 +70,13 @@ export const defaultOptions: HTMLReactParserOptions = {
...
-						width={ width }
-						className={ className }
+						height={ height || 0 }
+						width={ width || 0 }
+						className={ className || '' }
 						fill={ shouldFill }
-						style={ styleObject }
+						style={ styleObject ?? {} }
 						image={ imageAttributes }
 					/>
```
justlevine, line 81: **Never _ever_ default to a garbage value** Unless you want open a new tab with nowURL want to display an image of size `0`, or want a classname of `&nbsp;`

Either we (in order of preference)
- shouldn't set the prop
- should set the prop to `null` or `undefined`, and add it to the type
- Cannot be trusted to use `exactOptionalPropertyType` and should close this PR w.o merging.

cc: @ayushnirwal (in case you think there's a grey area i'm not accounting for)

packages/next/src/template-renderer/index.tsx
```diff
@@ -27,8 +27,13 @@ export async function TemplateRenderer( {
 	const headerList = await headers(); // headers() returns a Promise from NextJS 19.
 	const pathname = headerList.get( 'x-current-path' );
 
-	const { editorBlocks, bodyClasses, stylesheets, scripts, scriptModules } =
-		await getTemplateData( pathname || '/' );
+	const {
+		editorBlocks,
+		bodyClasses,
+		stylesheets = [],
```
justlevine, line 33: Why?

packages/query/src/utils/parse-template.ts
```diff
@@ -46,11 +46,12 @@ export default function parseQueryResult(
...
-		editorBlocks: parseEditorBlocks( templateByUri ),
-		scripts: parseEnqueuedScripts( templateByUri, wordpressUrl ),
-		scriptModules: parseScriptModules( templateByUri, wordpressUrl ),
-		bodyClasses: parseBodyClasses( templateByUri ),
+		stylesheets:
+			parseEnqueuedStylesheets( wordpressUrl, templateByUri ) || [],
+		editorBlocks: parseEditorBlocks( templateByUri ) || [],
+		scripts: parseEnqueuedScripts( templateByUri, wordpressUrl ) || [],
+		scriptModules: parseScriptModules( templateByUri, wordpressUrl ) || [],
+		bodyClasses: parseBodyClasses( templateByUri ) || [],
```
justlevine, line 54: 1. Why do we need to coerce ? (This feels like an issue with our parser functions)
2. We do we need to coerce _to an empty array_ (vs `null` or whatever).

justlevine (changes requested):
Lot of things I like here, but a lot of coercions that I don't think I understand. 

As per usual, I only flagged the first instance of a repeat problem.

---
PS:

1. I updated the PR title per Conventional Commits. Even though we use changesets for changelogs, git history is still important for people to navigate (ps: also why we squash PRs)
2. I marked the PR as `ready for review`, since my understanding from today's sync is that it is ready for _review_, it just might possibly not be ready to be merged. Try and use `draft PRs` for things you're still _actively working on_ , and use the PR description to indicate things like merge blockers, PR dependencies, etc.
(Draft or normal PR, you still can assign individuals for review and dismiss stale/re-ping them for another review)

packages/blocks/src/blocks/core-audio.tsx
BhumikP: yupp, will change code to follow same .

packages/next/src/react-parser/options.tsx
BhumikP: understood, will take this order preference and change code accordingly.

packages/next/src/template-renderer/index.tsx
BhumikP: removed

packages/next/src/components/link.tsx
BhumikP: Issue :
 [Ref](https://github.com/vercel/next.js/issues/50561) next/link tag is throwing error to add undefined type if type is not available. 
eg.   Types of property 'onMouseEnter' are incompatible.
      Type 'MouseEventHandler<HTMLAnchorElement> | undefined' is not assignable to type 'MouseEventHandler<HTMLAnchorElement>'. 
     
as a solution i was following [this](https://github.com/vercel/next.js/issues/36800) .
One another solution i found is to change the [InternalLinkProps](https://github.com/vercel/next.js/pull/57406).

 i need to look out for a better solution . @ayushnirwal  can help me here to find out better solution .

packages/blocks/src/blocks/core-file.tsx
```diff
@@ -78,6 +78,7 @@ const CoreFile: CoreFileType = ( {
 				<Link
 					href={ href }
 					className="wp-block-file__button wp-element-button"
+					//@ts-ignore allowing download prop in link
```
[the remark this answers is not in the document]
BhumikP: It is having issue linked to [Link](https://github.com/rtCamp/snapwp/pull/90#discussion_r1971051069)
Once that comment resolved , we can remove  ts-ignore

packages/blocks/src/blocks/core-audio.tsx
```diff
@@ -37,7 +35,12 @@ const CoreAudio: CoreAudioType = ( {
 	const classNames = cn( classNamesFromString );
 
 	return (
-		<figure className={ classNames } style={ styleObject }>
+		<figure
+			className={ classNames }
+			{ ...( style && {
+				style: getStylesFromAttributes( { style } ),
+			} ) }
```
justlevine, line 42: It's bad enough we gotta mix logic with our markup to avoid these `undefined`s. But I think it's still important to define `const styleObject` outside our `return ()`

justlevine: (I'm assuming it's scope creep to fix the return types on `getStylesFromAttributes()` ?)

packages/next/src/components/link.tsx
justlevine: (Based on the refs and some quick codebase dives), I think we want to explicitly add `| undefined` to the props where we want to treat an explicitly `undefined` prop the same as one which is `unset`, for `InternalLinkProps` and anywhere else that makes sense.

I defer to @ayushnirwal 🙇

packages/next/src/template-renderer/index.tsx
```diff
@@ -33,7 +33,7 @@ export async function TemplateRenderer( {
 	// @todo: Script modules should load before styles to handle ordering properly
 	return (
 		<>
-			<TemplateHead stylesheets={ stylesheets } />
+			<TemplateHead { ...( stylesheets && { stylesheets } ) } />
```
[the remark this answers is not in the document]
BhumikP: we can make default null stylesheets when getting from  `getTemplateData()` .

packages/query/src/utils/parse-template.ts
BhumikP: removed []

packages/blocks/src/blocks/core-audio.tsx
ayushnirwal: @BhumikP here is an idea.

```Typescript
// A map to hold all optional html attributes
const optionalProps:Record<string,unknown> = {}

const style = getStylesFromAttributes({styleAttributes});

if ( styleObject ){
   optionalProps.style = style
}

return (
  <figure
    className={classNames}
    //... any other attributes that WILL have Truthy values like classNames
   {...optionalProps}
</figure
```

packages/next/src/components/link.tsx
ayushnirwal: Yes. add a `| undefined` to `InternalLinkProps`

packages/next/src/template-renderer/index.tsx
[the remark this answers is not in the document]
justlevine: Per @ayushnirwal's [reply](https://github.com/rtCamp/snapwp/pull/90#discussion_r1973547969) we should _also_ add `| undefined` here

packages/next/src/components/link.tsx
BhumikP: Need to look into how can i change core next/link types

packages/types/src/global/link.ts
justlevine, line 1: This whole file/approach makes me nervous about tech debt.

Can we instead just extend the `NextImage` e.g. along the lines of:

```ts
type ExtendedImageProps = {
    [P in keyof ImageProps]?: ImageProps[P] | undefined;
};
```

cc @ayushnirwal

justlevine (changes requested):
Tl;dr

   - For passing props: whenever possible, we should avoid inlining logic inside `<Components ... />`, and instead assign that logic to a variable.
   - For our types: we should add `| undefined` to **every single optional type**  unless there is an explicit reason to treat `?:` different from an explicit `undefined`.
   - For _our_ components: if there's a value that we need to coerce to a different fallback shape to satisfy a type, it's likely a sign that we need to fix the type of either the prop or the variable.
   - Always use ` propName={ ...( var && { propName: var }` over `propName={ varName || undefined }`, since this rule means explicitly undefined props are _not_ the same as optional ones.

PS: Don't forget to check previously unresolved discussions before requesting re-review.

packages/core/src/config/snapwp-config-manager.ts
```diff
@@ -78,6 +78,7 @@ const defaultConfig: Partial< SnapWPEnv & SnapWPConfig > = {
  *
  * @return The configuration object.
  */
+//@ts-ignore - ignore check for nextUrl,homeUrl to run missing environment variable test.
```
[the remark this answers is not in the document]
BhumikP: It didn't caught, even our checks are successfull here.
Reason is commented are ignored in check.

packages/types/src/type-helpers/optional-props.ts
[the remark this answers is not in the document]
BhumikP: 1. Okay, we should go with typescript patterns, name it is  ```src/utils.ts``` file and if in future , we have any new type defs , we can add it in it.
2.  Sure, will add a doc-block in the file

packages/blocks/src/blocks/core-audio.tsx
BhumikP: Sure, we can add ``` | undefined ``` in getStylesFromAttributes's props and make it work with minimal change in code and avoid inline logic 
```
const styleObject = getStylesFromAttributes( { style } );
<figure className={ classNames } style={ styleObject }>
```

packages/blocks/src/blocks/core-paragraph.tsx
```diff
@@ -28,7 +28,7 @@ const CoreParagraph: CoreParagraphType = ( {
 		textColor,
 	} = attributes || {};
 
-	const styleObject = getStylesFromAttributes( { style } );
+	const styleObject = style ? getStylesFromAttributes( { style } ) : {};
```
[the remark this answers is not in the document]
BhumikP: we are attaching some values to styleObject, so if it is not available, we are creating empty one  with {} .

ayushnirwal (dismissed):

packages/next/src/template-renderer/template-scripts.tsx
```diff
@@ -16,11 +16,7 @@ const ScriptMap = ( { scripts }: { scripts: EnqueuedScriptProps[] } ) => (
 	<>
 		{ scripts?.map( ( { handle, src, ...rest }, id ) => {
 			return (
-				<Script
-					key={ handle || id }
-					src={ src || undefined }
-					{ ...rest }
-				/>
+				<Script key={ handle || id } src={ src || '' } { ...rest } />
```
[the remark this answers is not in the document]
justlevine: :-/ It's _our_ component and **we control the types**, so we can make `src` accept whatever we want it to.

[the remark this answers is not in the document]
BhumikP: fixed in [fc6db14](https://github.com/rtCamp/snapwp/pull/90/commits/fc6db149effae8c48ae6f402d5d88f072e2fad00)

packages/next/src/template-renderer/template-scripts.tsx
```diff
@@ -18,7 +18,7 @@ const ScriptMap = ( { scripts }: { scripts: EnqueuedScriptProps[] } ) => (
 			return (
 				<Script
 					key={ handle || id }
-					src={ src || undefined }
+					{ ...( src && { src } ) }
```
justlevine, line 21: Since `src` now takes `undefined` there's no need to spread it anymore.

packages/next/src/react-parser/options.tsx
```diff
@@ -71,13 +70,19 @@ export const defaultOptions: HTMLReactParserOptions = {
...
+						} ) }
+						{ ...( height && { height } ) }
+						{ ...( width && { width } ) }
+						className={ className || '' }
 						fill={ shouldFill }
-						style={ styleObject }
+						{ ...( style && {
+							style: getStyleObjectFromString( { style } ),
+						} ) }
 						image={ imageAttributes }
```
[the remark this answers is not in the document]
justlevine: Bump

justlevine (changes requested):
Personally I use the GitHub Pull Requests extension for VSCode, but if you're on browser and have a doubt about what still needs to be done, a good place to start is the `Conversations` button.

image

packages/next/src/template-renderer/template-scripts.tsx
BhumikP: Here we will  have to go with ```src={ src || undefined }``` , 
as changing ```src={src}``` , we will need to change props type and props here are `EnqueuedScriptProps[]`
which is being used somewhere else to match with graphql generated types .
so we will need to changes those ```__generated``` types to get it work .

justlevine: Nothing in that reply is accurate. [We control the types](https://github.com/rtCamp/snapwp/blob/c3c045c4868a335da5f382414d41aa50d191d745/packages/next/src/components/script.tsx#L11).

Cc @ayushnirwal

packages/next/src/react-parser/options.tsx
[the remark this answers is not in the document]
BhumikP: resolved

packages/blocks/src/blocks/core-paragraph.tsx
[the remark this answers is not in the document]
justlevine: Handled in 8ad82fcc1e539506dc0c7c30b7d2f94e04cb9329

packages/blocks/src/blocks/core-video.tsx
```diff
@@ -110,7 +113,9 @@ const CoreVideo: CoreVideoType = ( {
 				src={ src }
 				playsInline={ playsInline || undefined }
 			>
-				<Tracks tracks={ formattedTracks } />
+				<Tracks
+					{ ...( formattedTracks && { tracks: formattedTracks } ) }
```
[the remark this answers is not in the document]
justlevine: Handled in 8ad82fcc1e539506dc0c7c30b7d2f94e04cb9329

justlevine (changes requested):
I got sick of repeating myself so I handled the open comments in 166fb4f172ec27c86bd8f977c3b57a85200d3c46

Only things remaining are:
- @ayushnirwal should we be adding `| undefined` to _all_ of our undefined types in `@core/props` or `@types`? (IMO yes, since unless _we_ differentiate on behavior, we should absorb the headache for end-users who also want the `exactOptionalPropertyType` enforced locally, but I defer to you).
- _after_ a decision/change per the above (so the package diff is accurate), we need to `npm run changeset`

ayushnirwal:
>@ayushnirwal should we be adding | undefined to all of our undefined types in @core/props or @types? (IMO yes, since unless we differentiate on behavior, we should absorb the headache for end-users who also want the exactOptionalPropertyType enforced locally, but I defer to you).

Most of the types do not need `| undefined` according to my understanding. I see no cases where we need to set a prop (to a block renderer) to be undefined. It can be just left undefined. The same logic is applied to type describing env vars.
Other remaining types (related to block manager) may need `| undefined` but they are not fully "baked" yet. Decisions for them can be made when we do that.

.changeset/tender-jobs-itch.md
```diff
@@ -0,0 +1,9 @@
+---
+"@snapwp/blocks": patch
+"@snapwp/query": patch
+"@snapwp/types": patch
+"@snapwp/core": patch
+"@snapwp/next": patch
+---
+
+refctor: Add TS exactOptionalPropertyType rule
```
justlevine, line 9:
```suggestion
refactor: Enforce `exactOptionalPropertyType` TypeScript rule
```

PR #88, opened 2025-02-21
feat: Add eslint-comments/recommended rules

Pathan-Amaankhan:
## What
- This PR adds the following comment rulesets by extending `eslint-comments/recommended`:

| Rule | Description |
| :-- | :-- |
| [eslint-comments/disable-enable-pair](https://mysticatea.github.io/eslint-plugin-eslint-comments/rules/disable-enable-pair.html) | require a `eslint-enable` comment for every `eslint-disable` comment |
| [eslint-comments/no-aggregating-enable](https://mysticatea.github.io/eslint-plugin-eslint-comments/rules/no-aggregating-enable.html) | disallow a `eslint-enable` comment for multiple `eslint-disable` comments |
| [eslint-comments/no-duplicate-disable](https://mysticatea.github.io/eslint-plugin-eslint-comments/rules/no-duplicate-disable.html) | disallow duplicate `eslint-disable` comments |
| [eslint-comments/no-unused-enable](https://mysticatea.github.io/eslint-plugin-eslint-comments/rules/no-unused-enable.html) | disallow unused `eslint-enable` comments |

*The above rulesets reference is taken from [official eslint-plugin-eslint-comments documentation](https://mysticatea.github.io/eslint-plugin-eslint-comments/rules/).

### Related Issue(s):
- https://github.com/rtCamp/headless/issues/250

## Testing Instructions
- Run `npm install` && `npm run lint`.

## Additional Info
- This PR should be reviewed after https://github.com/rtCamp/snapwp/pull/84 is reviewed/merged & the base branch is updated (to remove any non-required changesets).
- As this PR fixes the codebase, we won't be getting any `eslint` errors.

## Checklist

-   [x] I have read the [Contribution Guidelines](https://github.com/rtCamp/snapwp/blob/develop/.github/CONTRIBUTING.md).
-   [x] My code is tested to the best of my abilities.
-   [x] My code passes all lints (ESLint, tsc, prettier etc.).
-   [x] My code has detailed inline documentation.
-   [ ] I have added unit tests to verify the code works as intended.
-   [ ] I have updated the project documentation as needed.
-   [ ] I have added a changeset for this PR using `npm run changeset`.

PR #87, opened 2025-02-21
dev: Implement `noPropertyAccessFromIndexSignature` tsconfig rule

BhumikP:
## What

Implemention of ```"noPropertyAccessFromIndexSignature": true``` rule

## Why

To make our ts rules more strict

### Related Issue(s):

https://github.com/rtCamp/headless/issues/317

## Checklist

-   [x] I have read the [Contribution Guidelines](https://github.com/rtCamp/snapwp/blob/develop/.github/CONTRIBUTING.md).
-   [x] My code is tested to the best of my abilities.
-   [x] My code passes all lints (ESLint, tsc, prettier etc.).
-   [ ] My code has detailed inline documentation.
-   [ ] I have added unit tests to verify the code works as intended.
-   [ ] I have updated the project documentation as needed.
-   [ ] I have added a changeset for this PR using `npm run changeset`.

PR #91, opened 2025-02-21
feat: Add `dot-notation` eslint rule

Pathan-Amaankhan:
## What
- This PR will add [dot-notation](https://eslint.org/docs/latest/rules/dot-notation) eslint ruleset. This ruleset forces the use of dot notation wherever possible.
- The `{ allowKeywords: false }` option added to ruleset will prevent use of reserved keywords to be used in dot notation. (ref: https://eslint.org/docs/latest/rules/dot-notation#options)

### Related Issue(s):

## Testing Instructions

## Additional Info
- This PR should be reviewed after https://github.com/rtCamp/snapwp/pull/88 is reviewed/merged & the base branch is updated (to remove any non-required changesets).

## Checklist


.eslintrc.cjs
```diff
@@ -5,6 +5,7 @@ module.exports = {
 	},
 	extends: '@snapwp/eslint-config',
 	parser: '@typescript-eslint/parser',
+	plugins: [ '@stylistic/jsx', '@typescript-eslint', 'jest' ],
```
justlevine, line 8: Remove this leak

.eslintrc.cjs
```diff
@@ -48,6 +49,9 @@ module.exports = {
 			},
 		],
 
+		// Enforce the use of dot notation over square brackets.
+		'dot-notation': [ 'error', { allowKeywords: false } ],
```
justlevine, line 53: Why aren't we using the ts version of this rule? https://github.com/rtCamp/snapwp/pull/87/files#r1968399510

Pathan-Amaankhan: This is a rule that should be added to our package.

TS version of this rule requires using a parser that we won't have/ship in our package.

---

https://github.com/rtCamp/snapwp/pull/87/files#r1968399510

In case of `noPropertyAccessFromIndexSignature`, we should use `@typescript-eslint/dot-notation` but only for our monorepo.

By default `noPropertyAccessFromIndexSignature` throws error during `npm run build` if we use dot notation for an object key whose type is not defined.

Eg:

```TypeScript
interface Car {
   wheel: number;
}

let car1: Car = {
    wheel: 4,
    brand: 'abc'
};

console.log( car1.wheel ); // `noPropertyAccessFromIndexSignature` will not give error as wheel has its type defined in Car interface.
console.log( car1.brand ); // `noPropertyAccessFromIndexSignature` will give error as brand does not have its type defined in Car interface.
console.log( car1['brand']); // This will not throw any error as per `noPropertyAccessFromIndexSignature` as `noPropertyAccessFromIndexSignature` enforces use of square bracket for keys not having type defined but on the other hand as per core eslint dot-notation ruleset, we should be using `.brand` instead of `['brand']`. So, here arrises a conflict in rules.
```

To prevent the conflict explained in the above example, we use `@typescript-eslint/dot-notation` for TS files as it explicitly handles the above case and allows the use of square brack without conflicting with `noPropertyAccessFromIndexSignature`.

As `noPropertyAccessFromIndexSignature` is configured in tsconfig, this issue will only occure in TypeScript files.

Pathan-Amaankhan:
- This PR is rebased to develop.
- Updated PR: https://github.com/rtCamp/snapwp/pull/99

PR #95, opened 2025-02-23
dev: add compat for ScriptModuleImportTypeEnum

justlevine:
## What

This PR adds compatibility for `ScriptModuleDependency.importType` now returning an enum instead of a string.

## Why

Compatibility with https://github.com/rtCamp/snapwp-helper/pull/88

## How

 `toUpperCase()` coersion is used for backwards-compatibility. This will be stripped for our next breaking release (when we drop support for the snapwp-helper v0.1.0

## Additional Info

I didn't bother narrowing down the types. it feels like there's gonna be a lot of type changes and no need to bump them now unnecessarily, imo.

## Checklist

### DIFF
--- a/.changeset/poor-lands-study.md
+++ b/.changeset/poor-lands-study.md
@@ -1,0 +1,5 @@
+---
+"@snapwp/next": patch
+---
+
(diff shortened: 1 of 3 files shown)


ashutoshgautams (approved):
LGTM 🚀 

TY. 🙇🏻

PR #102, opened 2025-02-25
chore: Add `jsdoc/require-param-type` eslint rule

Pathan-Amaankhan:
## What
- This PR enforces parameter type in function documentation by adding [jsdoc/require-param-type](https://github.com/gajus/eslint-plugin-jsdoc/blob/main/.README/rules/require-param-type.md) eslint rule.

### Related Issue(s):

## Testing Instructions

## Additional Info

## Checklist

### DIFF
--- a/packages/blocks/src/blocks/core-cover.tsx
+++ b/packages/blocks/src/blocks/core-cover.tsx
@@ -21,7 +21,7 @@ const DEFAULT_FOCAL_POINT = { x: 0.5, y: 0.5 };
  *
  * @see https://github.com/WordPress/gutenberg/blob/887243c1cb2b3280934fbff05d4af8e46a3feddc/packages/block-library/src/cover/shared.js#L27
  *
- * @param focalPoint - The focal point coordinates object containing x and y values.
+ * @param {FocalPoint} focalPoint The focal point coordinates object containing x and y values.
  *
  * @return CSS position string or undefined if no focal point
  */
@@ -36,12 +36,14 @@ const mediaPosition = ( focalPoint?: FocalPoint | null ): string => {
 
 /**
  * Renders the core/cover block.
- * @param root0 - The component props object
- * @param root0.attributes - Block attributes including styling and media settings
- * @param root0.children - Child elements to render inside the cover
- * @param root0.renderedHtml - Pre-rendered HTML string for class extraction
- * @param root0.connectedMediaItem - The connected media item object
- * @param root0.mediaDetails - The media details object
+ *
+ * @param {Object}                            root0                    The component props object
+ * @param {CoreCoverProps.attributes}         root0.attributes         Block attributes including styling and media settings
+ * @param {ReactNode}                         root0.children           Child elements to render inside the cover
+ * @param {CoreCoverProps.renderedHtml}       root0.renderedHtml       Pre-rendered HTML string for class extraction
+ * @param {CoreCoverProps.connectedMediaItem} root0.connectedMediaItem The connected media item object
+ * @param {CoreCoverProps.mediaDetails}       root0.mediaDetails       The media details object
+ *
  * @return The rendered cover block or null if using featured image
  */
 const CoreCover: CoreCoverType = ( {
--- a/packages/blocks/src/blocks/core-cover.tsx
+++ b/packages/blocks/src/blocks/core-cover.tsx
@@ -37,12 +37,12 @@ const mediaPosition = ( focalPoint?: FocalPoint | null ): string => {
 /**
  * Renders the core/cover block.
  *
- * @param {Object}                            root0                    The component props object
- * @param {CoreCoverProps.attributes}         root0.attributes         Block attributes including styling and media settings
- * @param {ReactNode}                         root0.children           Child elements to render inside the cover
- * @param {CoreCoverProps.renderedHtml}       root0.renderedHtml       Pre-rendered HTML string for class extraction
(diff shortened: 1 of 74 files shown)


packages/blocks/src/blocks/core-cover.tsx
```diff
@@ -36,12 +36,14 @@ const mediaPosition = ( focalPoint?: FocalPoint | null ): string => {
...
- * @param root0.connectedMediaItem - The connected media item object
- * @param root0.mediaDetails - The media details object
+ *
+ * @param {Object}                            root0                    The component props object
+ * @param {CoreCoverProps.attributes}         root0.attributes         Block attributes including styling and media settings
+ * @param {ReactNode}                         root0.children           Child elements to render inside the cover
+ * @param {CoreCoverProps.renderedHtml}       root0.renderedHtml       Pre-rendered HTML string for class extraction
+ * @param {CoreCoverProps.connectedMediaItem} root0.connectedMediaItem The connected media item object
+ * @param {CoreCoverProps.mediaDetails}       root0.mediaDetails       The media details object
+ *
```
justlevine, line 46: I think we make this `props` instead of `root0` for consistency w. the others

packages/blocks/src/editor-blocks-renderer/index.tsx
```diff
@@ -11,9 +11,11 @@ type EditorBlocksRendererProps = {
...
 /**
  * A react component to render editor blocks.
- * @param props - Props.
- * @param props.editorBlocks - A list of blocks to be rendered.
- * @param props.blockDefinitions - Blocks rendering functions.
+ *
+ * @param {Object}                                 props                  Props.
+ * @param {EditorBlocksRendererProps.editorBlocks} props.editorBlocks     A list of blocks to be rendered.
+ * @param {EditorBlocksRendererProps}              props.blockDefinitions Blocks rendering functions.
+ *
```
justlevine, line 18:
```suggestion
 * @param {EditorBlocksRendererProps.editorBlocks}     props.editorBlocks     A list of blocks to be rendered.
 * @param {EditorBlocksRendererProps.blockDefinitions} props.blockDefinitions Blocks rendering functions.
 *
```

packages/core/src/logger/index.ts
```diff
@@ -14,8 +14,8 @@ export enum LOGTYPE {
 /**
  * Logs a message to the console.
  *
- * @param type - The type of log message.
- * @param args - The arguments to log.
+ * @param {LOGTYPE} type The type of log message.
```
justlevine, line 17: We need a rule to block any "ts-only" features like `enum`. `LOGTYPE` should be a union of string literals.

packages/core/src/utils/find-element-and-get-class-names.ts
```diff
@@ -3,8 +3,8 @@ import { getClassNamesFromString, cn } from '@/utils';
 /**
  * Extracts the class names from the specified HTML element within the rendered HTML string.
  *
- * @param renderedHtml - The rendered HTML string.
- * @param elementSelector - The element selector (class, id, or tag name) If selector is not provided then it will select the first element and will return it's classNames.
+ * @param {string|null} renderedHtml    The rendered HTML string.
```
justlevine, line 6: `| undefined` ?

justlevine (changes requested):
PS: Is it a me thing that I'm not getting type-hinting for our param types - even _after_ building and reloading VSCode?

image

justlevine:
> @justlevine Do we wanna treat params the same as return types. Types via TS, Description via JsDoc?

@ayushnirwal IMO [unlike return types](https://rtcamp.slack.com/archives/CK2CEH4F6/p1741612957695869?thread_ts=1741609184.163489&cid=CK2CEH4F6) params are rarely as self-documenting, which means:

- We always want to require a @param _description_
- If we're already requiring a @param , then imo we should make sure the annotation is "complete" with a doc-type so users don't need to look at both the definition and the signature.

packages/core/src/utils/find-element-and-get-class-names.ts
Pathan-Amaankhan: Resolved in https://github.com/rtCamp/snapwp/pull/102/commits/60e5dbdf04282386e341ad66c35bc473b6e98dc7

packages/core/src/logger/index.ts
Pathan-Amaankhan: Hi @justlevine,
Can you please elaborate a bit. 
`enum` is TypeScript specific type & `LOGTYPE` is itself enum, but if we use it in TS files, it should not cause any problems.

Pathan-Amaankhan: Updated in https://github.com/rtCamp/snapwp/pull/102/commits/9252280fd0cfbd0ee9d2a2d892da6901bc589d12.

justlevine: 👏union👏of👏string👏literals👏

justlevine (changes requested):
I've got nothing to add beyond [this](https://github.com/rtCamp/snapwp/pull/102#issuecomment-2710913173) .

Add a changeset after @ayushnirwal reviews/approves (so the package diff will be correct), then ping me to squash-merge.

ayushnirwal:
> PS: Is it a me thing that I'm not getting type-hinting for our param types - even after building and reloading VSCode?
@Pathan-Amaankhan Is this resolved? It's happening on my machine too.

justlevine:
Closed in favor of #142

PR #103, opened 2025-02-26
refactor!: Change `location` to `groupLocation` for script placement enum type

ashutoshgautams:
## What
Change `location` property to `groupLocation` with proper enum type to match backend GraphQL schema updates

## Why
This PR updates the frontend code to be compatible with the backend changes that switched from using `location: String` to `groupLocation: ScriptLoadingGroupLocationEnum`

### Related Issue(s):
- Part of: rtCamp/snapwp

## How
- Created a `ScriptLoadingGroupLocationEnum` type with 'HEADER' and 'FOOTER' values
- Updated prop interfaces to use `groupLocation` instead of `location`
- Updated component logic to use the new enum values
- Modified GraphQL query to fetch `groupLocation` instead of `location`
- Updated documentation to reflect the new property name

## Testing Instructions
1. Add a script with `groupLocation="HEADER"` in a component
2. Verify script loads in the header with 'beforeInteractive' strategy
3. Add a script with `groupLocation="FOOTER"` in a component
4. Verify script loads in the footer with 'afterInteractive' strategy
5. Check that the GraphQL query correctly retrieves scripts with their locations

## Additional Info
This change includes a changeset file (`poor-lands-study.md`) to document the patch version update.

## Checklist
- [x] I have read the [Contribution Guidelines](https://github.com/rtCamp/snapwp/blob/develop/.github/CONTRIBUTING.md).
- [x] My code is tested to the best of my abilities.
- [x] My code passes all lints (ESLint, tsc, prettier etc.).
- [x] My code has detailed inline documentation.
- [ ] I have added unit tests to verify the code works as intended.
- [x] I have updated the project documentation as needed.
- [x] I have added a changeset for this PR using `npm run changeset`.

### DIFF
--- a/.changeset/fluffy-crabs-rhyme.md
+++ b/.changeset/fluffy-crabs-rhyme.md
@@ -1,7 +1,7 @@
 ---
-"@snapwp/query": patch
-"@snapwp/core": patch
-"@snapwp/next": patch
+"@snapwp/query": minor
+"@snapwp/core": minor
+"@snapwp/next": minor
 ---
 
-Refactor: Change location to groupLocation for script placement enum type
+refactor!: Change `location` to `groupLocation` for script placement enum type
--- a/.changeset/fluffy-crabs-rhyme.md
+++ b/.changeset/fluffy-crabs-rhyme.md
@@ -1,0 +1,7 @@
+---
+"@snapwp/query": patch
+"@snapwp/core": patch
+"@snapwp/next": patch
+---
+
+Refactor: Change location to groupLocation for script placement enum type
--- a/packages/core/src/props/index.ts
+++ b/packages/core/src/props/index.ts
@@ -24,15 +24,12 @@ export type StyleSheetProps = {
 	handle?: string | null;
 };
 
-// Define enum for script loading locations to match backend changes
-export type ScriptLoadingGroupLocationEnum = 'HEADER' | 'FOOTER';
-
 export type EnqueuedScriptProps = {
 	after?: ( string | null )[] | null;
 	before?: ( string | null )[] | null;
 	extraData?: string | null;
 	handle?: string | null;
-	groupLocation?: ScriptLoadingGroupLocationEnum | null;
+	groupLocation?: 'HEADER' | 'FOOTER' | null;
 	src?: string | null;
 	loadingStrategy?: 'ASYNC' | 'DEFER' | null;
 	version?: string | null;
--- a/packages/core/src/props/index.ts
+++ b/packages/core/src/props/index.ts
@@ -24,12 +24,16 @@ export type StyleSheetProps = {
 	handle?: string | null;
 };
 
+// Define enum for script loading locations to match backend changes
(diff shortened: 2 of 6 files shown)


packages/core/src/props/index.ts
```diff
@@ -24,12 +24,16 @@ export type StyleSheetProps = {
...
+export type ScriptLoadingGroupLocationEnum = 'HEADER' | 'FOOTER';
+
 export type EnqueuedScriptProps = {
 	after?: ( string | null )[] | null;
 	before?: ( string | null )[] | null;
 	extraData?: string | null;
 	handle?: string | null;
-	location?: string | null;
+	// Updated from 'location' to 'groupLocation'
+	groupLocation?: ScriptLoadingGroupLocationEnum | null;
```
justlevine, line 36: Comment is unnecessary, as is an enum right now

packages/next/README.md
```diff
@@ -137,7 +137,7 @@ export default function ScriptExample() {
                 src="https://example.com/main.js"
                 handle="example-main-script"
                 loadingStrategy="ASYNC"
-                location="header"
+                groupLocation="header"
```
justlevine, line 140: Should be uppercased value now that it's an enum

.changeset/fluffy-crabs-rhyme.md
```diff
@@ -0,0 +1,7 @@
+---
+"@snapwp/query": patch
```
justlevine, line 2: This is a breaking change, so all three should be minor

.changeset/fluffy-crabs-rhyme.md
```diff
@@ -0,0 +1,7 @@
...
+"@snapwp/core": patch
+"@snapwp/next": patch
+---
+
+Refactor: Change location to groupLocation for script placement enum type
```
justlevine, line 7:
```suggestion
refactor!: Change `location` to `groupLocation` for script placement enum type
```

packages/core/src/props/index.ts
justlevine:
```suggestion
	groupLocation?: 'HEADER' | 'FOOTER' | null;
```

packages/next/src/components/script.tsx
```diff
@@ -1,13 +1,16 @@
...
+export type ScriptLoadingGroupLocationEnum = 'HEADER' | 'FOOTER';
+
 interface ScriptInterface {
 	after?: ( string | null )[] | null;
 	before?: ( string | null )[] | null;
 	extraData?: string | null;
 	handle?: string | null;
 	loadingStrategy?: 'ASYNC' | 'DEFER' | null;
-	location?: string | null;
+	groupLocation?: ScriptLoadingGroupLocationEnum | null;
```
justlevine, line 13:
```suggestion
	groupLocation?: 'HEADER' | 'FOOTER' | null;
```

justlevine (changes requested):
Once the enum is inline fully and the type removed, this should be good to merge.

justlevine (approved):
Breaking change, so holding off merge for now.

PR #105, opened 2025-02-26
chore: Add `react/jsx-boolean-value` eslint rule to codebase

Pathan-Amaankhan:
## What
- This PR implements [react/jsx-boolean-value](https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/jsx-boolean-value.md) forcing to omit the true value passed to JSX component's prop to maintain consistency in code. (JFI: if a prop requires a boolean value & if no value is passed, by default it is true [[Ref](https://www.freecodecamp.org/news/react-props-cheatsheet/#:~:text=React%20props%20passed%20with%20just%20their%20name%20have%20a%20value%20of%20true)])

### Related Issue(s):

## Testing Instructions

## Additional Info

## Checklist

### DIFF
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -70,6 +70,8 @@ module.exports = {
 
 		// Restrict the use of empty functions.
 		'no-empty-function': 'error',
+
+		'react/jsx-boolean-value': 'error',
 	},
 	overrides: [
 		{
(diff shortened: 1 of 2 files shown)


justlevine (commented):
I'm 100% for enforcing a pattern, but its wholly subjective whether we `always` or `never` require an explicit `true`.

Leaving it to @ayushnirwal to decide

ayushnirwal:
> I'm 100% for enforcing a pattern, but its wholly subjective whether we always or never require an explicit true.

I'm more inclined toward never requiring an explicit True.

justlevine (approved):

PR #108, opened 2025-02-27
dev: Add `import/default` to `.js`, `.jsx`, `.ts` & `.tsx` files

Pathan-Amaankhan:
## What
- This PR enforces [import/default](https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/default.md) rule on `.js`, `.jsx`, `.ts` & `.tsx` files by removing `"import/default": off`.
- As per this rule, if a default import is requested, the rule will report if there is no default export in the imported module.

### Related Issue(s):

## Testing Instructions

## Additional Info

## Checklist

### DIFF
--- a/.changeset/swift-paths-change.md
+++ b/.changeset/swift-paths-change.md
@@ -1,0 +1,8 @@
+---
+"@snapwp/blocks": patch
+"@snapwp/types": patch
+"@snapwp/core": patch
+"@snapwp/next": patch
+---
+
+Update jsx compiler
--- a/.changeset/swift-paths-change.md
+++ b/.changeset/swift-paths-change.md
@@ -5,4 +5,4 @@
 "@snapwp/next": patch
 ---
 
-Update jsx compiler
+chore: Enforce "import/default" eslint ruleset.
--- a/packages/blocks/src/blocks/core-audio.tsx
+++ b/packages/blocks/src/blocks/core-audio.tsx
@@ -1,4 +1,3 @@
-import * as React from 'react';
 import {
 	cn,
 	getClassNamesFromString,
--- a/packages/blocks/src/blocks/core-audio.tsx
+++ b/packages/blocks/src/blocks/core-audio.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import * as React from 'react';
 import {
 	cn,
 	getClassNamesFromString,
--- a/packages/blocks/src/blocks/core-button.tsx
+++ b/packages/blocks/src/blocks/core-button.tsx
@@ -1,4 +1,4 @@
-import { type ButtonHTMLAttributes } from 'react';
+import type { ButtonHTMLAttributes } from 'react';
 import { cn, getStylesFromAttributes, replaceHostUrl } from '@snapwp/core';
 import { getConfig } from '@snapwp/core/config';
 import { Link, Parse } from '@snapwp/next';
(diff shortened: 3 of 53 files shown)


packages/blocks/src/blocks/core-audio.tsx
```diff
@@ -1,4 +1,4 @@
-import React from 'react';
+import * as React from 'react';
```
justlevine, line 1: Do we even need to still import `React` in v17+?

I really don't like importing _all_ of react (versus what gets transpiled as the "default"

packages/blocks/src/blocks/core-button.tsx
```diff
@@ -1,4 +1,5 @@
-import React, { type ButtonHTMLAttributes } from 'react';
+import * as React from 'react';
+import { type ButtonHTMLAttributes } from 'react';
```
justlevine, line 2: **If** ([big if](https://github.com/rtCamp/snapwp/pull/108/files#r1973924580) we need `import * as React`, then make the entire second statement typed, to ensure bundlers dont get confused.

```suggestion
import type { ButtonHTMLAttributes } from 'react';
```

packages/blocks/src/blocks/core-audio.tsx
ayushnirwal: @Pathan-Amaankhan This should not be the case. `import React from 'react'` should work with `allowSyntheticDefaultImports` and `esModuleInterop` switched to true. 
You need to fix it with [parserOptions](https://typescript-eslint.io/packages/parser/) in `eslint.cjs`

justlevine:
(also needs a changeset)

packages/blocks/src/blocks/core-audio.tsx
Pathan-Amaankhan: Hi @ayushnirwal,
Adding [parserOptions](https://typescript-eslint.io/packages/parser/) does not fix the issue.

ayushnirwal: Try `jsx`:`react-jsx` in package's TS config. We may be using the wrong environment.

Pathan-Amaankhan: Yup, updating `jsx: react` to `jsx: react-jsx` in compilerOptions of tsconfig.base.json resolved the issue. After reading, found that compiling JSX using `react` outputs `.js` files with JSX changed to the equivalent React.createElement calls ([ref](https://www.typescriptlang.org/tsconfig/#jsx)) which is now deprecated ([ref](https://react.dev/reference/react/createElement)). 
Due to this, we were forced to use older syntax.

Resolved this in https://github.com/rtCamp/snapwp/pull/108/commits/19bdb83b46f090ac7bac17117447d45e12a7fa25

packages/blocks/src/blocks/core-button.tsx
Pathan-Amaankhan: Resolved this in https://github.com/rtCamp/snapwp/pull/108/commits/32f637f5588d860d5533c571cad5a2c1ee6745bb

ayushnirwal (approved):

.changeset/swift-paths-change.md
```diff
@@ -0,0 +1,8 @@
+---
+"@snapwp/blocks": patch
+"@snapwp/types": patch
+"@snapwp/core": patch
+"@snapwp/next": patch
+---
+
+Update jsx compiler
```
justlevine, line 8: 1. Conventional Commits
2. What the PR actually does. (Not sure how this relates to jsx compiling).


```suggestion
chore: Enforce "import/default" eslint ruleset.
```

PR #111, opened 2025-03-02
feat: robust handling for internal uri

SH4LIN:
## What

- This PR will allow developers to individually specify the home URL and site URL.
- This PR will add wrapper functions to identify internal URLs more robustly
- This PR will replace the internal URL with the NextJS app URL using JS URL API instead of simple string replace.

## Why

- Current approach for the replace is prone to bugs.

### Related Issue(s):

- Part of https://github.com/rtCamp/headless/issues/413

## How

- We will parse the URL using the URL JS API and replace the properties with the nextApp url properties and build the URL again from it.

## Checklist


.env.example
```diff
@@ -4,8 +4,10 @@
 # The headless frontend domain URL. Uncomment this line and ensure the value matches the URL used by your frontend app.
 # NEXT_PUBLIC_URL=http://localhost:3000
 
+NEXT_PUBLIC_WORDPRESS_SITE_URL=https://headless-demo.local
```
justlevine, line 7: If we're renaming these anyway maybe we shorten `_WORDPRESS_` to `_WP_`?

examples/nextjs/starter/.env.example
```diff
@@ -1,14 +1,16 @@
 # Enable if connecting to a self-signed cert
-# NODE_TLS_REJECT_UNAUTHORIZED="0"
+# NODE_TLS_REJECT_UNAUTHORIZED=0
 
-# The headless frontend domain URL
-NEXT_PUBLIC_URL="http://localhost:3000"
+# The headless frontend domain URL. Uncomment this line and ensure the value matches the URL used by your frontend app.
+# NEXT_PUBLIC_URL=http://localhost:3000
```
justlevine, line 5: Did you mean to revert uncommenting `NEXT_PUBLIC_URL`?

ayushnirwal (changes requested):

docs/config-api.md
```diff
@@ -18,7 +18,8 @@ SnapWP uses the following `.env` variables to configure your Next.js app.
...
 | --------------------------------------- | -------- | ---------------------------------------- | --------------------------------------------------------------------------------- | -------------------------- |
 | `NEXT_PUBLIC_URL`                       | Yes      |                                          | The URL of the Next.js site.                                                      | `nextUrl`                  |
-| `NEXT_PUBLIC_WORDPRESS_URL`             | Yes      |                                          | The WordPress frontend domain URL.                                                | `homeUrl`                  |
+| `NEXT_PUBLIC_WORDPRESS_HOME_URL`        | Yes      |                                          | The WordPress frontend domain URL.                                                | `homeUrl`                  |
+| `NEXT_PUBLIC_WORDPRESS_SITE_URL`        | Yes      |                                          | The WordPress site URL.                                                           | `siteUrl`                  |
```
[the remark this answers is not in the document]
justlevine: https://developer.wordpress.org/reference/functions/home_url/

https://developer.wordpress.org/reference/functions/site_url/

.env.example
justlevine: `NEXT_PUBLIC_NEXT_APP_URL` is redundant but you could do `NEXT_PUBLIC_FRONTEND_URL` if you want (assuming @ayushnirwal has no issues).

`NEXT_PUBLIC_WP_GRAPHQL_ENDPOINT` - seems this just has 3 extra characters. As there's only 1 GraphQL endpoint in use, how does the `_WP_` help with self documentation?

examples/nextjs/starter/.env.example
SH4LIN: Nope, I made changes to snapwp/.env.example and copied those changes to /snapwp/examples/nextjs/starter/.env.example

.env.example
SH4LIN: That aligns with all the other WP dependent variables. Like `NEXT_PUBLIC_WP_SITE_URL`, `NEXT_PUBLIC_WP_HOME_URL`, `NEXT_PUBLIC_WP_UPLOADS_PATH`, `NEXT_PUBLIC_WP_REST_URL_PREFIX`

justlevine: Because, unlike `GRAPHQL_ENDPOINT` , if you remove `WP`, they're not self documenting
(But yeah I'm not happy w. The name of the last one either)

ayushnirwal: @justlevine Do you have any suggestion for new names?

ayushnirwal (dismissed):
LGTM.

The env vars for CI/CD are needed to be updated.

justlevine:
(@justlevine audit env var names)

.env.example
justlevine: Need more thought on this but I think `NEXT_PUBLIC_USE_CORS_PROXY` should be removed and instead using `NEXT_PUBLIC_CORS_PROXY_PREFIX: string|false` . Default value when unset would be `is development env ? '/proxy' : false`

Then `useCorsProxy` would be renamed to `hasCorsProxy`

Also:
- `nextUrl` => `frontendUrl`
- `NEXT_PUBLIC_WP_UPLOADS_PATH` vs `uploadsDirectory` - its either a path or a directory.

justlevine: Also wondering if our `siteUrl` and `homeUrl` should be prefixed with `wp`. Sure, unlike env files we're not worried about the potential for a user-defined var to cause confusion (since these only come from `getConfig()`), but I still wonder how self-documenting `siteUrl` and `homeUrl` are on the frontend when those are not the _frontend's_ site or home url but are specific WP terminology.

Same reason btw why I think we should _remove_ `WP` from the graphql/rest endpoints and make them `NEXT_PUBLIC_GRAPHQL_ENDPOINT` and `NEXT_PUBLIC_REST_URL_PREFIX`: these only need to be explicitly identified as `WP` if a user decides they need another graphql or rest endpoint in the app _and_ they suck at naming things themselves to prevent confusion.

packages/core/src/utils/to-internal-uri.ts
justlevine, line 1: Why does this function get its own file but `url-checker.ts` has 3 (none of which are named `url-checker`)

packages/core/src/utils/to-internal-uri.ts
```diff
@@ -0,0 +1,32 @@
...
+import { isInternalUrl } from '@/utils/url-checker';
+
+/**
+ * Convert a URL to an internal URL.
+ *
+ * @param url The URL to convert.
+ *
+ * @return The internal URL.
+ */
+export const toInternalUrl = ( url: string ): string => {
```
justlevine, line 11: The whole point is is to convert this into a UR*I* not a UR*L*. a relative `/` that can be consumed by `next/link`

packages/core/src/utils/to-internal-uri.ts
```diff
@@ -0,0 +1,32 @@
...
+
+	const urlObject = new URL( url );
+	const nextUrlObject = new URL( nextUrl );
+
+	nextUrlObject.pathname = urlObject.pathname;
+	nextUrlObject.search = urlObject.search;
+	nextUrlObject.hash = urlObject.hash;
+	nextUrlObject.password = urlObject.password;
+	nextUrlObject.username = urlObject.username;
+
```
justlevine, line 30: There's no way this is sustainable. Why aren't we just string-replacing?

packages/core/src/utils/url-checker.ts
```diff
@@ -0,0 +1,38 @@
+import { getConfig } from '@/config';
+
```
justlevine, line 2: 1. Relative UR**I** not absolute UR**L**
2. We either want to outright ignore the protocol and just check the `homeUrl`/`siteUrl` _domains_, or we want an `ignoreProtocol` (or similar) param that defaults to ignoring it.

packages/core/src/utils/url-checker.ts
```diff
@@ -0,0 +1,38 @@
...
+ * Check if a URL is the home URL.
+ *
+ * @param url The URL to check.
+ *
+ * @return Whether the URL is the home URL.
+ */
+export const isWPHomeUrl = ( url: string ): boolean => {
+	const { homeUrl } = getConfig();
+	return (
+		url.startsWith( '/' ) || url === homeUrl || url.startsWith( homeUrl )
```
justlevine, line 13: `url.startsWith( '/' )` should `isInternalUri()` true not `isWPHomeUri()`

packages/core/src/utils/to-internal-uri.ts
justlevine: Separately, thinking we rename this `toFrontendUri()` to match the updated ENV var.

packages/next/src/components/link.tsx
```diff
@@ -33,22 +33,13 @@ export default function Link( {
...
+	const { graphqlEndpoint } = getConfig();
 
 	const internalUri = href
-		? replaceHostUrl(
-				href,
-				homeUrl,
-				nextUrl
-				// @todo: Remove replace when the graphql endpoint is removed from the pagination links.
-		  )?.replace( `/${ graphqlEndpoint }`, '' )
+		? toInternalUrl( href )?.replace( `/${ graphqlEndpoint }`, '' )
```
justlevine, line 39: Restore the todo about stripping the graphql endpoint from pagination links

examples/nextjs/starter/.env.example
```diff
@@ -1,20 +1,28 @@
 # Enable if connecting to a self-signed cert
-# NODE_TLS_REJECT_UNAUTHORIZED="0"
+# NODE_TLS_REJECT_UNAUTHORIZED=0
 
 # The headless frontend domain URL
-NEXT_PUBLIC_URL="http://localhost:3000"
+NEXT_PUBLIC_FRONTEND_URL=http://localhost:3000
+
+NEXT_PUBLIC_WP_SITE_URL=https://headless-demo.local
```
justlevine, line 7: Disabled by default, and unset.

```suggestion
# NEXT_PUBLIC_WP_SITE_URL=
```

examples/nextjs/starter/.env.example
```diff
@@ -1,20 +1,28 @@
...
 
 # The WordPress "frontend" domain URL
-NEXT_PUBLIC_WORDPRESS_URL="http://localhost"
+NEXT_PUBLIC_WP_HOME_URL=https://headless-demo.local
```
justlevine, line 11:
```suggestion
# The WordPress "frontend" domain URL e.g. https://my-headless-site.local
# NEXT_PUBLIC_WP_HOME_URL=

```

examples/nextjs/starter/.env.example
justlevine, line 1: We should Az sort _our_ variables (ignoring `NEXT_PUBLIC_` prefix

justlevine (changes requested):

justlevine:
(Also, dont forget to add a changelog)

packages/core/src/utils/to-internal-uri.ts
SH4LIN: toInternalUri is not a check action. And url-checker is logical categorisation of function, under url-checker there are functions which checks/compares the URL whether those are Home URL or site url or internal url. This categorisation is for internal. NextJS devs or even our other packages can import `isWPHomeUrl` from `@snapwp/core`

packages/core/src/utils/to-internal-uri.ts
SH4LIN: > There's no way this is sustainable

Why?

packages/core/src/utils/to-internal-uri.ts
SH4LIN: @justlevine I have a doubt, if my Home url is `headless.local` and my site url is `headless.local/wordpress` now when `headless.local/page-1` is in url we will replace it with `/page-1` should we do the same for siteUrl? like replace `headless.local/wordpress/xyz.png` with `/xyz.png`

ayushnirwal (dismissed):

packages/core/src/utils/to-internal-uri.ts
justlevine: I asked you "why" first ;-) 

Because you're specifically allow-listing and mapping individual object properties, which means we need to maintain this list, and there's no way for the user to.

packages/core/src/utils/to-internal-uri.ts
justlevine: Thats a _horrible_ non-answer. 

1. file names should be self documenting to what they contain.
2. file names should file the same conventions.

Tell me how any of the files or functions in `to-frontend-uri`, `url-slash-modify` or `url-checker` meet either of these conditions?

image

If you or @ayushnirwal cant think of an acceptable naming convention, LMK - i'm happy to put this on hold until I have time.

packages/core/src/utils/to-internal-uri.ts
justlevine: Reverse the logic and you should be fine - e.g. (pretending we don't care about the protocol, though as already flagged we [_very much do_](https://github.com/rtCamp/snapwp/pull/111/#discussion_r1980278230)

```
const domainToReplace = isWPSiteUrl( url ) ? siteUrl : homeUrl;

// pretend we're doing something that wont break if there's a different protocol
return addLeadingSlash( url.replace( domainToReplace, '' ) );
```

packages/core/src/utils/url-checker.ts
justlevine: Bump

examples/nextjs/starter/.env.example
```diff
@@ -1,20 +1,25 @@
...
+# The WordPress "frontend" domain URL e.g. https://my-headless-site.local
+NEXT_PUBLIC_WP_HOME_URL=
 
 # The WordPress REST URL Prefix
-# NEXT_PUBLIC_WORDPRESS_REST_URL_PREFIX=/wp-json
+# NEXT_PUBLIC_REST_URL_PREFIX=/wp-json
 
-# Token used for authenticating GraphQL introspection queries
-INTROSPECTION_TOKEN=
+# NEXT_PUBLIC_WP_SITE_URL=
```
justlevine, line 22:
```suggestion
# The WordPress "backend" Site Address. Uncomment if different than `NEXT_PUBLIC_WP_HOME_URL` e.g. https://my-headless-site.local/wp/ 
# NEXT_PUBLIC_WP_SITE_URL=
```

packages/blocks/src/blocks/core-button.tsx
```diff
@@ -1,11 +1,11 @@
...
+import { cn, getStylesFromAttributes } from '@snapwp/core';
 import { Link, Parse } from '@snapwp/next';
 import type {
 	CoreButton as CoreButtonType,
 	CoreButtonProps,
 } from '@snapwp/types';
+import type { ButtonHTMLAttributes } from 'react';
+
 
 /**
```
justlevine, line 10:
```suggestion
} from '@snapwp/types';
import type { ButtonHTMLAttributes } from 'react';

/**
```

packages/core/src/config/snapwp-config-manager.ts
```diff
@@ -27,14 +29,10 @@ export interface SnapWPEnv {
...
 	restUrlPrefix: string;
-	/**
-	 * URL prefix for WP assets loaded from 'wp-includes' dir . Defaults to `/proxy`.
-	 */
-	corsProxyPrefix: string;
 	/**
 	 * Flag to enable cors middleware which proxies assets from WP server.
 	 */
-	useCorsProxy?: boolean;
+	hasCorsProxy?: string | false;
```
justlevine, line 35: Must not have been clear in my previous comment.

The _exported_ config should have a `corsProxyPrefix?: string | undefined` _and_ a `hasCorsProxy:bool`.

The difference is that that `hasCorsProxy = !! corsProxyPrefix` instead of needing to be defined/managed by a separate ENV var.

packages/core/src/config/snapwp-config-manager.ts
```diff
@@ -6,13 +6,15 @@ import type { HTMLReactParserOptions } from 'html-react-parser';
...
 	 */
-	nextUrl: string;
+	frontendUrl: string;
 	/**
-	 * The home URL of the WordPress site. Defaults to `process.env.NEXT_PUBLIC_WORDPRESS_URL`.
-	 *
-	 * TODO: Update variable name from homeUrl to siteUrl. ref: https://github.com/rtCamp/headless/pull/272#discussion_r1907601796
+	 * The site URL of the WordPress site. Defaults to `process.env.NEXT_PUBLIC_WP_SITE_URL`.
+	 */
+	siteUrl: string;
```
justlevine, line 15: As previously flagged, imo `getConfig().siteUrl` and `getConfig().homeUrl` are _not_ self documenting, and should be changed to `wpSiteUrl` and `wpHomeUrl` respectively.

The others are fine, because their names don't conflict with frontend semantics.

packages/core/src/config/snapwp-config-manager.ts
```diff
@@ -80,13 +77,16 @@ const defaultConfig: Partial< SnapWPEnv & SnapWPConfig > = {
...
 const envConfig = (): Partial< SnapWPEnv > => ( {
 	/* eslint-disable n/no-process-env -- These are the env variables we want to manage. */
-	nextUrl: process.env.NEXT_PUBLIC_URL,
-	homeUrl: process.env.NEXT_PUBLIC_WORDPRESS_URL,
 	graphqlEndpoint: process.env.NEXT_PUBLIC_GRAPHQL_ENDPOINT,
-	uploadsDirectory: process.env.NEXT_PUBLIC_WORDPRESS_UPLOADS_PATH,
-	restUrlPrefix: process.env.NEXT_PUBLIC_WORDPRESS_REST_URL_PREFIX,
-	useCorsProxy: process.env.NEXT_PUBLIC_USE_CORS_PROXY === 'true',
-	corsProxyPrefix: process.env.NEXT_PUBLIC_CORS_PROXY_PREFIX,
+	hasCorsProxy: process.env.NEXT_PUBLIC_USE_CORS_PROXY,
```
justlevine, line 81: https://github.com/rtCamp/snapwp/pull/111#discussion_r1987710723

```suggestion
	hasCorsProxy: !! process.env.NEXT_PUBLIC_CORS_PROXY_PREFIX,
```

packages/types/src/global/env.ts
```diff
@@ -6,11 +6,12 @@ declare global {
...
+			NEXT_PUBLIC_FRONTEND_URL?: string;
 			NEXT_PUBLIC_USE_CORS_PROXY?: string;
-			NEXT_PUBLIC_WORDPRESS_REST_URL_PREFIX?: string;
-			NEXT_PUBLIC_WORDPRESS_UPLOADS_PATH?: string;
-			NEXT_PUBLIC_WORDPRESS_URL?: string;
+			NEXT_PUBLIC_WP_SITE_URL?: string;
+			NEXT_PUBLIC_REST_URL_PREFIX?: string;
+			NEXT_PUBLIC_WP_UPLOADS_DIRECTORY?: string;
+			NEXT_PUBLIC_WP_HOME_URL?: string;
 			NODE_ENV?: 'development' | 'production' | 'test';
```
justlevine, line 15: This needs to be sorted and updated.

packages/core/src/utils/to-internal-uri.ts
SH4LIN: @justlevine How about I create an individual file with the name of the function following kebab case?

justlevine: Unless @ayushnirwal thinks one-util-per-file is an _ideal_ pattern in and of itself, I'd prefer to (hold if needed and) at least attempt to implement a cohesive naming convention before we (imo) increase the cognitive complexity of having to juggle multiple files.

SH4LIN: Then we should hold it until we come up with better names/naming conventions.

packages/core/src/utils/url-checker.ts
SH4LIN: @justlevine Let me know current changes addresses this feedback

packages/core/src/config/snapwp-config-manager.ts
```diff
@@ -6,35 +6,37 @@ import type { HTMLReactParserOptions } from 'html-react-parser';
 
 export interface SnapWPEnv {
 	/**
-	 * The URL of the Next.js site. Defaults to `process.env.NEXT_PUBLIC_URL`.
+	 * URL prefix for WP assets loaded from 'wp-includes' dir . Defaults to `/proxy`.
 	 */
-	nextUrl: string;
+	corsProxyPrefix: string;
```
justlevine, line 11: No?
```suggestion
	corsProxyPrefix?: string | undefined;
```

packages/core/src/config/snapwp-config-manager.ts
```diff
@@ -63,12 +65,12 @@ type ConfigSchema< T > = {
  * Default configuration.
  */
 const defaultConfig: Partial< SnapWPEnv & SnapWPConfig > = {
+	corsProxyPrefix: '/proxy',
```
justlevine, line 68:
```suggestion
	corsProxyPrefix: process.env.NODE_ENV === 'development' ? '/proxy' : undefined,
```

packages/core/src/config/snapwp-config-manager.ts
```diff
@@ -63,12 +65,12 @@ type ConfigSchema< T > = {
...
 	graphqlEndpoint: 'index.php?graphql',
-	uploadsDirectory: '/wp-content/uploads',
 	restUrlPrefix: '/wp-json',
-	corsProxyPrefix: '/proxy',
+	uploadsDirectory: '/wp-content/uploads',
 	// eslint-disable-next-line n/no-process-env -- We're using `NODE_ENV` to derive a default value.
-	useCorsProxy: process.env.NODE_ENV === 'development',
+	hasCorsProxy: true,
```
justlevine, line 73: If I understand the code correctly, we dont need a default since it's inherited.

```suggestion
```

packages/core/src/config/snapwp-config-manager.ts
```diff
@@ -6,35 +6,37 @@ import type { HTMLReactParserOptions } from 'html-react-parser';
 
 export interface SnapWPEnv {
```
justlevine, line 7: Sort A-z

packages/core/src/utils/url-checker.ts
```diff
@@ -0,0 +1,73 @@
...
+		return false;
+	}
+	const { wpHomeUrl } = getConfig();
+
+	if ( ignoreProtocol ) {
+		return new URL( url ).host === new URL( wpHomeUrl ).host;
+	}
+
+	return url === wpHomeUrl || url.startsWith( wpHomeUrl );
+};
```
justlevine, line 26: Witespace + comments.

```suggestion
	if ( ! isValidUrl( url ) ) {
		return false;
	}

	const { wpHomeUrl } = getConfig();

	if ( ignoreProtocol ) {
		// Compare just the domains.
		return new URL( url ).host === new URL( wpHomeUrl ).host;
	}

	// Compare with the protocol.
	return url === wpHomeUrl || url.startsWith( wpHomeUrl );
};
```

packages/core/src/config/snapwp-config-manager.ts
SH4LIN: I think we should keep it's value like that, making it undefined is same as having hasCorsProxy: false.

justlevine (commented):
These are comments from an _old_ code review. Not sure why they didn't get shared w. The rest, nor if they're still accurate 🤦‍♂️

I'll re-rereview tomorrow night

packages/core/src/config/snapwp-config-manager.ts
justlevine: Per [slack](https://rtcamp.slack.com/archives/CK2CEH4F6/p1741952816843009?thread_ts=1740982755.056769&cid=CK2CEH4F6), `hasCorsProxy` can be removed if it provides no DX value. Programmatically, it is just as representable as by ` !! corsProxyPrefix` (and fewer sources of truth is generally a good thing)

packages/core/src/utils/url-checker.ts
```diff
@@ -0,0 +1,73 @@
...
+	if ( ! isValidUrl( url ) ) {
+		return false;
+	}
+
+	return (
+		url.startsWith( '/' ) ||
+		isWPHomeUrl( url, ignoreProtocol ) ||
+		isWPSiteUrl( url, ignoreProtocol )
+	);
+};
```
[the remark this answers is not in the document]
SH4LIN: Reverting my change as well if url `/` starts with this then URL API will throw error making it invalid URL

packages/core/src/utils/to-frontend-uri.ts
```diff
@@ -0,0 +1,26 @@
...
+		return url;
+	}
+
+	if ( url.startsWith( '/' ) ) {
+		return url;
+	}
+
+	const { wpHomeUrl, wpSiteUrl } = getConfig();
+
+	const domainToUpdate = isWPHomeUrl( url ) ? wpHomeUrl : wpSiteUrl;
```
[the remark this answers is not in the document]
justlevine: You were worried about a use case where the site address is in a subdir: `mysite.com/wp`. to avoid it you should reverse the logic here and check for siteUrl and fallback to home url.

docs/config-api.md
```diff
@@ -15,16 +15,15 @@ SnapWP uses the following `.env` variables to configure your Next.js app.
...
-| `NEXT_PUBLIC_GRAPHQL_ENDPOINT`          | No       | `index.php?graphql`                      | The relative path to the WordPress GraphQL endpoint.                              | `graphqlEndpoint`          |
-| `NEXT_PUBLIC_WORDPRESS_UPLOADS_PATH`    | No       | `/wp-content/uploads`                    | The relative path to the WordPress uploads directory.                             | `uploadsDirectory`         |
-| `NEXT_PUBLIC_WORDPRESS_REST_URL_PREFIX` | No       | `/wp-json`                               | The WordPress REST API URL prefix.                                                | `restUrlPrefix`            |
-| `NEXT_PUBLIC_USE_CORS_PROXY`            | No       | `process.env.NODE_ENV === 'development'` | Whether to use a CORS proxy for the WordPress API.                                | `useCorsProxy`             |
-| `NEXT_PUBLIC_CORS_PROXY_PREFIX`         | No       | `/proxy`                                 | The prefix of the CORS proxy.                                                     | `corsProxyPrefix`          |
```
justlevine, line 27: We need to restore this

justlevine (changes requested):
Last round of feedback from me (hopefully)

If @ayushnirwal has no suggestions for a [utils naming convention](https://github.com/rtCamp/snapwp/pull/111#discussion_r1987862444), then I'll see what GPT recommends

ayushnirwal (approved):

justlevine (commented):
@Ta5r can you PTAL at the changes to the env variables (both documented and the comments that get output in the `.env.example`), and make a matching PR for `snapwp-helper`?

justlevine (approved):
Can be merged after @Ta5r updates the env vars in `rtcamp/snapwp-helper`

PR #112, opened 2025-03-03
feat: Display error fallback when no editor blocks

Swanand01:
## What

This PR displays an error when `editorBlocks` are not defined.

## Why

Currently, no message is displayed when editorBlocks is undefined.

### Related Issue(s):

https://github.com/orgs/rtCamp/projects/141/views/10?pane=issue&itemId=99060513

## How

The `DefaultError` component is used to display this message: 'Error: Unable to render content. `editorBlocks` is not defined. This may be due to missing template data or an issue with data fetching.'

## Screenshots

image

## Checklist

### DIFF
--- a/.changeset/empty-pigs-fry.md
+++ b/.changeset/empty-pigs-fry.md
@@ -1,0 +1,5 @@
+---
+"@snapwp/next": minor
+---
+
+feat: Display error fallback when there are no editor blocks
--- a/.changeset/empty-pigs-fry.md
+++ b/.changeset/empty-pigs-fry.md
@@ -1,5 +1,5 @@
 ---
-"@snapwp/next": minor
+"@snapwp/next": patch
 ---
 
 feat: Display error fallback when there are no editor blocks
--- a/packages/next/src/template-renderer/index.tsx
+++ b/packages/next/src/template-renderer/index.tsx
@@ -5,7 +5,6 @@ import { TemplateHead } from './template-head';
 import { TemplateScripts } from './template-scripts';
 import Script from 'next/script';
 import type { BlockData } from '@snapwp/types';
-import DefaultError from '@/components/default-error';
 
 export type TemplateRendererProps = {
 	getTemplateData?: ( typeof QueryEngine )[ 'getTemplateData' ];
@@ -31,6 +30,12 @@ export async function TemplateRenderer( {
 	const { editorBlocks, bodyClasses, stylesheets, scripts, scriptModules } =
 		await getTemplateData( pathname || '/' );
 
+	if ( ! editorBlocks ) {
+		throw new Error(
+			'Error: Unable to render content. `editorBlocks` is not defined. This may be due to missing template data or an issue with data fetching.'
+		);
+	}
+
 	// @todo: Script modules should load before styles to handle ordering properly
 	return (
 		<>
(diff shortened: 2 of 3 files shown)


packages/next/src/template-renderer/index.tsx
```diff
@@ -40,8 +41,17 @@ export async function TemplateRenderer( {
...
-						{ editorBlocks && children( editorBlocks ) }
+						{ editorBlocks ? (
+							children( editorBlocks )
+						) : (
+							<DefaultError
+								error={
+									new Error(
+										'Error: Unable to render content. `editorBlocks` is not defined. This may be due to missing template data or an issue with data fetching.'
+									)
+								}
```
ayushnirwal, line 52: Don't remove reset button. Refresh the page on click.

Swanand01: Next.js won't let me pass the refresh function from the `TemplateRenderer` to the `DefaultError` component, since the former is a server component and the latter is a client component. I tried creating a middle client component that passed `router.refresh` to `DefaultError`, but it didn't work either.

Thus, I am using `router.refresh` as a fallback for the reset prop if it is not passed.
https://github.com/rtCamp/snapwp/blob/108af92fea33be2bc94b85d1a63c62cdb0adb533/packages/next/src/components/default-error.tsx#L29-L33

SH4LIN, line 52: Why are we doing it like this? 
@ayushnirwal correct me if I am wrong.
If editorBlocks are not there, then there will be nothing to show on the front end right?

If that's the case just throw the error from the TemplateRenderer function, which will be caught in our errorBoundary, and it will display out defaultError component with it's own reset function passed from the nextJS.

With this approach if developer has customized there own error page then it won't use that.

SH4LIN (changes requested):

packages/next/src/template-renderer/index.tsx
ayushnirwal: Yes! Great Idea. @Swanand01 let's do that. 
In the Future we can also throw errors/warnings if other data (bodyClasses, stylesheets, scripts, scriptModules) don't match the expectation.

packages/next/src/template-renderer/index.tsx
```diff
@@ -30,6 +30,12 @@ export async function TemplateRenderer( {
 	const { editorBlocks, bodyClasses, stylesheets, scripts, scriptModules } =
 		await getTemplateData( pathname || '/' );
 
+	if ( ! editorBlocks ) {
```
SH4LIN, line 33: Our editorBlocks can be `BlockData[] | undefined` `if ( ! editorBlocks )` this condition will check for undefined but what if we are getting an empty array following suggestion will handle that edge case as well.

```suggestion
	if ( ! editorBlocks || ! editorBlocks.length ) {
```

SH4LIN (changes requested):

packages/next/src/template-renderer/index.tsx
```diff
@@ -30,7 +30,7 @@ export async function TemplateRenderer( {
 	const { editorBlocks, bodyClasses, stylesheets, scripts, scriptModules } =
 		await getTemplateData( pathname || '/' );
 
-	if ( ! editorBlocks ) {
+	if ( ! editorBlocks || editorBlocks.length === 0 ) {
```
justlevine, line 33: Nit: 
```suggestion
	if ( ! editorBlocks?.length ) {
```

ayushnirwal (approved):

SH4LIN (dismissed):

justlevine (approved):
My gut tells me this isn't very composable, and the error catch/fallback needs to happen _inside_ `children()` and leave the `<main><div>` wrapper in-tact.

_But_ this is an incremental improvement that doesn't really fit into our scope, so I'm merging as is.

@ayushnirwal please LMK if you think my feeling here is correct and if I should stub a draft issue to follow this up.

.changeset/empty-pigs-fry.md
```diff
@@ -0,0 +1,5 @@
+---
+"@snapwp/next": minor
```
justlevine, line 2: Changesets is dumb. the type refers to the digit-location _not_ actual semver 

```suggestion
"@snapwp/next": patch
```

justlevine (approved):

PR #117, opened 2025-03-04
refactor: allow default option in cli prompts

BhumikP:
## What

 When user do not enter an path for creating project,
Following things will be taken place:

  - Default folder for the project will be  ```snapwp-frontend``` 
  - .env file will be created from .env.example

### Related Issue(s):

https://github.com/rtCamp/headless/issues/410

## Checklist

### DIFF
--- a/.changeset/slimy-chicken-cross.md
+++ b/.changeset/slimy-chicken-cross.md
@@ -1,0 +1,5 @@
+---
+"snapwp": patch
+---
+
+Implement default cli option
--- a/packages/cli/src/snapwp.cjs
+++ b/packages/cli/src/snapwp.cjs
@@ -9,7 +9,7 @@ const path = require( 'path' );
 const readline = require( 'readline' );
 const { program } = require( 'commander' );
 
-const defaultEnvContent  = `
+const defaultEnvContent = `
 # Enable if connecting to a self-signed cert
 # NODE_TLS_REJECT_UNAUTHORIZED=0
 
--- a/packages/cli/src/snapwp.cjs
+++ b/packages/cli/src/snapwp.cjs
@@ -9,6 +9,29 @@ const path = require( 'path' );
 const readline = require( 'readline' );
 const { program } = require( 'commander' );
 
+const defaultEnvContent  = `
+# Enable if connecting to a self-signed cert
+# NODE_TLS_REJECT_UNAUTHORIZED=0
+
+# The headless frontend domain URL
+NEXT_PUBLIC_URL=
+
+# The WordPress "frontend" domain URL
+NEXT_PUBLIC_WORDPRESS_URL=
+
+# The WordPress GraphQL endpoint
+NEXT_PUBLIC_GRAPHQL_ENDPOINT=
+
+# The WordPress Uploads directory path
+# NEXT_PUBLIC_WORDPRESS_UPLOADS_PATH=/wp-content/uploads
+
+# The WordPress REST URL Prefix
+# NEXT_PUBLIC_WORDPRESS_REST_URL_PREFIX=/wp-json
+
+# Token used for authenticating GraphQL introspection queries
+INTROSPECTION_TOKEN=
+`;
+
 program
 	.option( '--proxy', 'Use proxy registry.' )
 	.option( '--use-defaults', 'Use default values.', false ) // Optional flag for using default value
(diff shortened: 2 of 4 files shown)


Pathan-Amaankhan (commented):
Few things observed:
1. When using `--use-defaults` flag Why are we waiting for user to click when everything is setup?

https://github.com/user-attachments/assets/330612db-27b3-450c-8de6-9ecebc90e682

2. As we are keeping the values empty for env variables when `--use-defaults` is used, we should be logging additional steps to setup env properly at the end.

3. Instead of using a flag i.e. --use-default, it will be good if we can show prompt and when left empty, a default directory will be selected (Same goes with env vars). This way we can omit `--use-defaults` flag entirely.
    - The prompt for directory can be: 
        ```
         Where would you like to create your new Headless WordPress frontend?
         Please enter a relative or absolute path (default: /User/../snapwp-frontend)
        ```
        Note: `/User/../snapwp-frontend` will be an absolute path.

Will be deferring to @justlevine.

cc: @ayushnirwal

packages/cli/src/snapwp.cjs
```diff
@@ -86,12 +112,20 @@ const openEditor = ( filePath ) => {
...
+		if ( ! options.useDefaults ) {
+			projectDir = await prompt(
+				'Thanks for using SnapWP!\n' +
+					'\nWhere would you like to create your new Headless WordPress frontend?\n' +
+					'Please enter a relative or absolute path: '
+			);
+		} else {
+			prompt(
+				'Thanks for using SnapWP!\n' +
+					`\nUsing default values , your project directory will be created at ${ projectDir } path.\n`
```
Pathan-Amaankhan, line 126:
```suggestion
					`\nUsing default values, your project directory will be created at ${ projectDir } path.\n`
```

Pathan-Amaankhan (changes requested):

.changeset/slimy-chicken-cross.md
```diff
@@ -0,0 +1,5 @@
+---
+"snapwp": patch
+---
+
+Implement default cli option
```
justlevine, line 5: https://github.com/rtCamp/snapwp/blob/c1ee5564e1046411dac0ccb5ace99c59fe337959/DEVELOPMENT.md?plain=1#L224-L228

justlevine (changes requested):
As defined in the parent task and mentioned previously on Slack there is 0 value in implementing a `--use-defaults` CLI _flag_ before we provide **prompt-level** defaults so the user can hit `enter` instead of manually typing in what is the most common happy-path.

Only **after** those prompt-level defaults are implemented, does it make sense to add the flag to holistically use those defaults without printing a prompt + requiring the using to hit `enter`.

(@Pathan-Amaankhan not specifically diving into your feedback because it's either recommending/including what's proscribed above, or obviated if the above is implemented correctly)

packages/cli/src/snapwp.cjs
```diff
@@ -9,7 +9,33 @@ const path = require( 'path' );
 const readline = require( 'readline' );
 const { program } = require( 'commander' );
 
-program.option( '--proxy', 'Use proxy registry.' ).parse();
+const defaultEnvContent = `
```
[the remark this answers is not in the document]
BhumikP: As ```.env``` creation depends on  user input,
So, we can simply create empty env file.

packages/cli/src/snapwp.cjs
```diff
@@ -86,12 +112,20 @@ const openEditor = ( filePath ) => {
...
+				'Thanks for using SnapWP!\n' +
+					'\nWhere would you like to create your new Headless WordPress frontend?\n' +
+					'Please enter a relative or absolute path: '
+			);
+		} else {
+			prompt(
+				'Thanks for using SnapWP!\n' +
+					`\nUsing default values, your project directory will be created at ${ projectDir } path.\n`
+			);
+		}
```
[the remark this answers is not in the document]
BhumikP: Can add some general instructions to setup env.
```'For setting up environment variables, please refer to the documentation at: https://github.com/rtCamp/snapwp/blob/b7c0472d95be624244ad2a5d01d4bcdaa29e91f3/packages/cli/README.md' ```

File: packages/cli/src/snapwp.cjs [hunk 5]
```diff
@@ -72,13 +72,24 @@ const openEditor = ( filePath ) => {
 
 ( async () => {
 	try {
+		let useDefaultPath = false;
+		let projectPath = './snapwp-frontend';
```
justlevine, line 76: This whole thing seems messy, can we instead handle defaults inside the prompt function - e.g `prompt( query, default )`?

BhumikP: Noted, will handle  default in prompt.
although we will need ```useDefaultPath``` to differentiate further that default prompt is being used.

justlevine:
Somehow we're _still_ going nowhere with this PR.

@BhumikP pick up (and test) #124 as a replacement for this one.

PR #119, opened 2025-03-05
feat: load favicon from WordPress server

SH4LIN:
## What

- This PR will load the favicon icon, apple touch icon and msapplication-tileImage from the WordPress server

### Related Issue(s):

- Part of: https://github.com/rtCamp/headless/issues/405

## How

- This PR will use the app router metadata API and using that metadata API and the icon data fetched from server it will add the necessary tags into the head

## Checklist


examples/nextjs/starter/src/app/layout.tsx
```diff
@@ -1,4 +1,5 @@
-import { RootLayout } from '@snapwp/next';
+import { RootLayout, generateRootMetaData } from '@snapwp/next';
```
ayushnirwal, line 1: Nit: I disagree with the notion of adding `root` to meta-data semantics. Just say metadata.

Why - I don't see [Nextjs's documentation](https://nextjs.org/docs/app/building-your-application/optimizing/metadata) using "root". Although they have used parent though in [this](https://arc.net/l/quote/thsbzuey) example. Let's omit using "root" like them.

ayushnirwal (changes requested):

examples/nextjs/starter/src/app/layout.tsx
SH4LIN: used `root` in the context of RootLayout that this metadata is generated from the RootLayout component (Metadata related to the RootLayout). When this function is imported in app/layout.tsx, there will be a generateMetaData function so we will need something to differentiate between that generateMetaData and our meta data from the root layout

ayushnirwal (dismissed):
LGTM

packages/next/src/root-layout/index.tsx
```diff
@@ -30,3 +32,132 @@ export async function RootLayout( {
...
+type Icon = {
+	sourceUrl: string;
+	height: string;
+	width: string;
+};
+
+type FormattedIcon = {
+	url: string;
+	sizes: string;
+};
```
ayushnirwal, line 51: Move types and interfaces to `@snapwp/types`

packages/next/src/root-layout/index.tsx
ayushnirwal, line 1: I think it is time to update the file structure. Leave `generateRootMetaData` in this file. Move `generateIconsMetaData` to its file with its utilities.

ayushnirwal (changes requested):

packages/types/src/queries/index.ts
justlevine, line 1: This is either counterintuitive _or_ indicative of a horrible antipattern hiding in this PR.

As a reminder, separation of concerns requires GraphQL Query (codegen-types) => our parsers/adapers => agnostic/composable frontend components.

IOW **only** types we should be distributing are the agnosting ones that our components expect. Further, the only reason for defining the types in the global types packages (versus colocated/directly `typeof ComponentProps`d) is to avoid circular dependency issues.

Tbe entire concept of a `types/queries` namespace hints to a coupling that's the exact opposite of the above behavior.

packages/next/src/root-layout/index.tsx
justlevine: Related (possible disagreement - didnt look into how the types are used, naming convention there was enough for me to 🚩 ): https://github.com/rtCamp/snapwp/pull/119#discussion_r1986430854

justlevine (changes requested):
Functionally working 🚀

Feedback on composability/code quality patterns below:

packages/next/src/root-layout/icons-metadata.ts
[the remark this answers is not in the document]
SH4LIN: I’m sorry, I’m a bit confused here. What should I do here? If nothing is Next.js specific, should I move it to another package? However, it wouldn’t be specific to that package either. The function that consumes this is specific to the Next.js Metadata API.

examples/nextjs/starter/src/app/layout.tsx
```diff
@@ -1,4 +1,5 @@
...
+import type { Metadata } from 'next';
 
 export default function Layout( { children }: { children: React.ReactNode } ) {
```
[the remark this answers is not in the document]
SH4LIN: That's not slipped in. The Next.js documentation uses this approach in their docs – [[Next.js Layouts](https://nextjs.org/docs/app/api-reference/file-conventions/layout#root-layouts)](https://nextjs.org/docs/app/api-reference/file-conventions/layout#root-layouts).  

Whoever implemented this followed the same approach.

packages/types/src/queries/index.ts
SH4LIN: Please correct me if I misunderstood anything.

We want our types to be as close to the component as possible, defining the exact type that the component expects. For the parser, we have decided not to modify anything; instead, we will pass the data down to the component and let it handle any nullish values. Correct?

examples/nextjs/starter/src/app/layout.tsx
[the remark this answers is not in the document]
justlevine: "Slipped through" means got passed code review; I assume that developers have a reason for _all_ their code decisions ;-)

If for some reason we cannot use `PropsWithChildren`, then:
1. `import type { ReactNode } from 'react';
2. Add a comment explaining why we cant use `PropsWithChildren`.

packages/types/src/queries/index.ts
justlevine: > We want our types to be as close to the component as possible, defining the exact type that the component expects. 

Correct.

> For the parser, we have decided not to modify anything; instead, we will pass the data down to the component and let it handle any nullish values.

Ummmm it depends?
We're creating an _agnostic_ component, the purpose of the parser is to coerce the GraphQL data into the shape expected by the component.
In general the reason to allow nullish values on the component props is so users can use them directly with minimal headaches so that's the theoretical "ideal shape", _not_ because any constraints on the parser.

packages/next/src/root-layout/icons-metadata.ts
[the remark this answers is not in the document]
justlevine: If there is value in a _user_ accessing these functions to do something outside of nextjs (favicons in cra?), then it make's sense to split it up. (cc @ayush.nirwal for final decision)

examples/nextjs/starter/src/app/layout.tsx
```diff
@@ -1,9 +1,23 @@
-import { RootLayout } from '@snapwp/next';
+import { RootLayout, generateRootMetaData } from '@snapwp/next';
+import type { Metadata } from 'next';
+import { PropsWithChildren } from 'react';
```
justlevine, line 3:
```suggestion
import type { PropsWithChildren } from 'react';
```

justlevine (changes requested):
Functionally, things are passing, but I would like @ayushnirwal 's feedback on the composability (generally, and per the open conversations on this diff).

Once that's done (since the package diff might change) `npm run changeset` and we should be able to merge this.

packages/next/src/root-layout/icons-metadata.ts
```diff
@@ -0,0 +1,140 @@
...
+		siteIcon: {
+			mediaItemUrl?: string;
+			mediaDetails: {
+				sizes?: IconData[];
+			};
+		};
+	};
+}
+
+type IconMetaData = {
```
ayushnirwal, line 14:
```suggestion
interface IconMetaData = {
```

packages/next/src/root-layout/icons-metadata.ts
```diff
@@ -0,0 +1,140 @@
...
+	faviconIcons: FormattedIconData[];
+	appleIcons: FormattedIconData[];
+	msApplicationTileIcon: IconData;
+};
+
+type IconData = {
```
ayushnirwal, line 20:
```suggestion
interface IconData = {
```

packages/next/src/root-layout/icons-metadata.ts
```diff
@@ -0,0 +1,140 @@
...
+	sourceUrl: string;
+	height: string;
+	width: string;
+};
+
+type FormattedIconData = {
```
ayushnirwal, line 26:
```suggestion
interface FormattedIconData = {
```

ayushnirwal (changes requested):
Although the patterns around queries, documents, client management are not ideal in the current state. I'm willing to merge this PR (after the last few feedbacks are addressed) for the following reasons. 

- It works.
- This adds parsers, some of the types around favicon and meta data management. When we implement the metadata management and data fetching management interfaces these will be used.

justlevine:
> Although the patterns around queries, documents, client management are not ideal in the current state. I'm willing to merge this PR 

@ayushnirwal can you flag some parts of the diff that deserve a `// @todo` and/or generally list what the next steps should theoretically be for improving things?

I don't want to merge this PR without making sure we have a (vague) plan to follow it up.

ayushnirwal:
```typescript
import type { DocumentNode, TypedQueryDocumentNode } from 'graphql';

export type Variables = Record< string, unknown >;

export type FetchQuery< T = unknown, V extends Variables = Variables > = (
	queryKey: string[],
	document: DocumentNode | TypedQueryDocumentNode< T, V >
) => Promise< T >;

export type getServerClient = (
	url: string,
	options: Record< string, unknown >
) => ServerClient;

export type getBrowserClient = (
	url: string,
	options: Record< string, unknown >
) => BrowserClient;

export interface ServerClient {
	fetchQuery: FetchQuery;
}

export interface BrowserClient {}

```

These types/interfaces are the general broad strokes for the composability in our data-fetching mechanisms. This allows users to 
- Implement their own ServerClient with the custom `getServerClient` and `fetchQuery`
- Compose their own ServerClient with the `getServerClient` and `fetchQuery` with implementations exported from `@snapwp/query`
- Use a pre-implemented ServerClient exported from `@snapwp/query`
- Any implementation of ServerClient will have the `fetchQuery` function to which a doc can be passed for querying.
- Wrappers will be built around `FetchQuery` to get functions like `FetchGlobalStyles`. These will be exported for internal or external use.

justlevine:
@ayushnirwal anything specific to favicons / site metadata (see the remaining open conversations on this PR) or just the underlying Query API?

ayushnirwal:
> Wrappers will be built around FetchQuery to get functions like FetchGlobalStyles. These will be exported for internal or external use.

- Similar to this we will create a wrapper `FetchFaviconDetails`
- `getIcons` (without the fetch call) and its internal utilities become data parsing utilities. Which allows us to move it to `@snapwp/core` along with the related types.  
- `FetchFaviconDetails` and renamed `getIcons` will be then composed to make `getMetaData` in `@snapwp/next`.
- The same pattern will be followed for the SEO-related meta data too.

packages/types/src/queries/index.ts
justlevine: Making as resolved since the file is gone.

packages/next/src/root-layout/icons-metadata.ts
```diff
@@ -0,0 +1,140 @@
...
+}
+
+interface FormattedIconData {
+	url: string;
+	sizes: string;
+}
+
+/**
+ * Fetch site icon data and filter them into categorized formats.
+ * @see - https://developer.wordpress.org/reference/functions/wp_site_icon/#:~:text=Displays%20site%20icon%20meta%20tags. Reason why we are different resolution icons into individual category.
```
justlevine, line 33: Documenting https://github.com/rtCamp/snapwp/pull/119#issuecomment-2736076355

```suggestion
 *
 * @see - https://developer.wordpress.org/reference/functions/wp_site_icon/#:~:text=Displays%20site%20icon%20meta%20tags. Reason why we are different resolution icons into individual category.
 *
 * @todo Refactor for composability alongside SEO metadata patterns 
```

packages/next/src/root-layout/index.tsx
```diff
@@ -30,3 +32,26 @@ export async function RootLayout( {
...
 	);
 }
+
+/**
+ * Generate and return root metadata, including icons and other metadata.
+ *
+ * @return Merged metadata.
+ */
+export async function generateRootMetaData(): Promise< Metadata > {
+	// Fetch icons in required format, apply faviconIcons and apple touch icons in icons metadata property while apply msapplication-TileImage in other metadata property
```
justlevine, line 42: Documenting https://github.com/rtCamp/snapwp/pull/119#issuecomment-2736076355

```suggestion
	/**
	 * Fetch icons in required format, apply faviconIcons and apple touch icons in icons metadata property while apply msapplication-TileImage in other metadata property.
	 *
	 * @todo Review composability when implementing SEO metadata
```

ayushnirwal (approved):

justlevine:
@SH4LIN what's with the failing tests?

Also can you add a changeset ?

(Once those are resolved this should be good to merge)

SH4LIN:
@justlevine Checks are passing now

packages/next/src/root-layout/icons-metadata.ts
```diff
@@ -36,19 +36,28 @@ interface FormattedIconData {
...
-		return {};
+		return {
+			faviconIcons: [],
+			appleIcons: undefined,
+			msApplicationTileIcon: undefined,
+		};
 	}
 
-	let fallBackIcons = {};
+	let fallBackIcons: IconMetaData = {
```
justlevine, line 51:
```suggestion
	let fallbackIcons: IconMetaData = {
```

.changeset/late-breads-listen.md
justlevine, line 1: Reminder, changesets is "dumb" and `major|minor|patch` is mapped blindly to `x.y.z`, not real semver

justlevine (approved):

PR #120, opened 2025-03-05
fix : Support area attribute as a fallback for `templatePartTagName` in `CoreTemplatePart`.

Ta5r:
## What

- This PR updates the `CoreTemplatePart` component to use the `area` attribute as a fallback when `templatePartTagName` is not provided.

## Why

- Ensures that the `area` attribute is available in the response from the GraphQL query.
- Allows the `CoreTemplatePart` component to use area as a fallback for `templatePartTagName`.
- Updates type definitions to include area, improving type safety and consistency.

### Related Issue(s):

- Fixes [Task: Add support for CoreTemplate block area_tag. #407](https://github.com/rtCamp/headless/issues/407)

## How

- Modified the logic to use area as a fallback for `templatePartTagName`.
- Ensured type safety by checking if area is a valid string before using it.
- Added area to the `CoreTemplatePartFrag` fragment.

## Testing Instructions

NA

## Screenshots

NA

## Additional Info

NA

## Checklist

### DIFF
--- a/packages/blocks/src/blocks/core-template-part.tsx
+++ b/packages/blocks/src/blocks/core-template-part.tsx
@@ -20,13 +20,17 @@ const CoreTemplatePart: CoreTemplatePartType = ( {
 	children,
 	attributes,
 }: CoreTemplatePartProps ) => {
-	const { templatePartTagName } = attributes || {};
-	/**
-	 * @todo add support for area_tag.
-	 * @see https://github.com/WordPress/gutenberg/blob/4775e7052b9e2ed7df46429e6e738de3faf2fb18/packages/block-library/src/template-part/index.php#L165
-	 */
-	const TagName = ( templatePartTagName ||
-		'div' ) as keyof JSX.IntrinsicElements;
+	const { templatePartTagName, area } = attributes || {};
+
+	let htmlTag = 'div';
+
+	if ( templatePartTagName ) {
+		htmlTag = templatePartTagName;
+	} else if ( typeof area === 'string' && area.trim() !== '' ) {
+		htmlTag = area;
+	}
+
+	const TagName = htmlTag as keyof JSX.IntrinsicElements;
 
 	/**
 	 * @todo replace with cssClassName once it's supported.
--- a/packages/types/src/blocks/props/core-template-part.ts
+++ b/packages/types/src/blocks/props/core-template-part.ts
@@ -3,6 +3,7 @@ import type { BaseAttributes, BaseProps } from '../base';
 
 export type CoreTemplatePartAttributes = BaseAttributes & {
 	templatePartTagName?: string;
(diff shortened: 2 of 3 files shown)


packages/blocks/src/blocks/core-template-part.tsx
```diff
@@ -20,13 +20,17 @@ const CoreTemplatePart: CoreTemplatePartType = ( {
...
-	 */
-	const TagName = ( templatePartTagName ||
-		'div' ) as keyof JSX.IntrinsicElements;
+	const { templatePartTagName, area } = attributes || {};
+
+	let htmlTag = 'div';
+
+	if ( templatePartTagName ) {
+		htmlTag = templatePartTagName;
+	} else if ( typeof area === 'string' && area.trim() !== '' ) {
```
ayushnirwal, line 29: Any reason not to do the same checks for `templatePartTagName`?

ayushnirwal (changes requested):

packages/blocks/src/blocks/core-template-part.tsx
justlevine: Better q: any reason to use a default value and a double conditional vs `htmlTag = templatePartTagName || area || 'div'` ?

(But either way we don't need to worry about trimming. )

.changeset/strong-weeks-punch.md
```diff
@@ -0,0 +1,7 @@
+---
+"@snapwp/blocks": minor
+"@snapwp/query": minor
+"@snapwp/types": minor
+---
+
+fix: Support area attribute as a fallback for templatePartTagName in CoreTemplatePart.
```
justlevine, line 7:
```suggestion
fix: Use `area` attribute as HTML tag fallback in `CoreTemplatePart` block component.
```

packages/types/src/blocks/props/core-template-part.ts
```diff
@@ -3,6 +3,7 @@ import type { BaseAttributes, BaseProps } from '../base';
 
 export type CoreTemplatePartAttributes = BaseAttributes & {
 	templatePartTagName?: string;
+	area?: string;
```
justlevine, line 6: Let's allow an explicit `| undefined` on both of these.

packages/cli/CHANGELOG.md
ayushnirwal, line 1: Why are there changes in this file?

justlevine: Because nobody's run `npm run format` yet

Ta5r: yes

justlevine (approved):
Updated the changeset, since it's a dumb check that doesn't understand semver semantics.

ayushnirwal (approved):

PR #123, opened 2025-03-13
feat: SEO support

Swanand01:
## What

This PR introduces the SEO feature, adding support for dynamic metadata handling, including Open Graph and Twitter metadata. 
It generates metadata for pages, posts, categories, and other entities.

## Why

SEO is essential for improving search visibility and social media sharing. This PR enables automatic metadata generation, ensuring pages have correct titles, descriptions, and images. It also resolves issues with metadata merging, particularly for the `/` route.

### Related Issue(s):

https://github.com/rtCamp/headless/issues/431

## Testing Instructions

1. Verify metadata for posts, pages, categories, and tags.
2. Check Open Graph and Twitter metadata on different routes.
3. Ensure global and route-specific metadata merge correctly.
4. Test homepage (/) metadata handling.

## Additional Info

- Uses GraphQL queries to fetch metadata dynamically.

## Checklist


Swanand01:
I have noticed a bug:
Refreshing or hard refreshing the page causes the meta tags to disappear from the `<head>` tag, and appear in the `<body>` tag instead. This is an issue with NextJS.
https://github.com/vercel/next.js/issues/46738
cc: @justlevine

justlevine:
> This is an issue with NextJS. [vercel/next.js#46738](https://github.com/vercel/next.js/issues/46738) 

@Swanand01 the issue you linked is closed with a matching PR...

Swanand01:
> > This is an issue with NextJS. [vercel/next.js#46738](https://github.com/vercel/next.js/issues/46738) 
> 
> 
> 
> @Swanand01 the issue you linked is closed with a matching PR...

Right, but the issue seems to persist. Also, it doesn't **always ** occur on refreshing/hard refreshing a page, but sometimes it does.

packages/core/src/logger/tests/logger.test.ts
```diff
@@ -17,6 +17,7 @@ describe( 'Logger', () => {
 	} );
 
 	it( 'should log a debug message', () => {
+		// @ts-ignore - Overriding NODE_ENV for testing logger behavior in different environments.
```
ayushnirwal, line 20: Put one on top of the file.

packages/query/src/queries/get-twitter-metadata.graphql
```diff
@@ -0,0 +1,31 @@
+query GetTwitterMetadata($uri: String!) {
+	nodeByUri(uri: $uri) {
```
justlevine, line 2: 1. We use `templateByUri`
2. Use fragments and interfaces `... on NodeWithFeaturedImage`
3. Every node should have the global `id` queried, so things are correctly cached by the fewhen possible.

packages/query/src/queries/get-route-metadata.graphql
```diff
@@ -0,0 +1,36 @@
+query GetRouteMetadata($uri: String!) {
+	nodeByUri(uri: $uri) {
```
justlevine, line 2: 1. `templateByUri`
2. These should be built up holistically with fragments + interfaces (`NodeWithTitle`, `NodeWithAuthor` etc)
3. use global `id`s on every node.

packages/next/src/seo/route/index.ts
```diff
@@ -0,0 +1,155 @@
...
+	parser: parseRouteSiteMetadata,
+};
+
+const DEFAULT_FETCHED_OBJECT = {};
+const DEFAULT_VALIDATED_OBJECT = {};
+const DEFAULT_PARSED_OBJECT = {};
+
+registerMetadataGenerator( OpenGraphMetaDataGenerator, 'open-graph' );
+registerMetadataGenerator( TwitterMetaDataGenerator, 'twitter' );
+registerMetadataGenerator( SiteMetaDataGenerator, 'site' );
```
justlevine, line 58: Maybe its the lack of docs on this file but I am not understanding the api, nor what\how the lifecycle happens or what functionality is pluggable.

packages/next/src/seo/utils/get-clean-excerpt.ts
```diff
@@ -0,0 +1,11 @@
+/**
+ * Removes all HTML tags from a given string.
+ *
+ * @param html - The HTML string to be stripped.
+ * @return The cleaned text without HTML tags.
+ */
+export default function getCleanExcerpt( html: string ): string {
```
justlevine, line 7: If we really need this function, probably better to call it `sanitizeHtml()` and to export it from `@/core`

packages/next/src/seo/root/site-metadata/types.ts
```diff
@@ -0,0 +1,8 @@
+/**
+ * Represents the Site metadata of a site.
+ */
+export interface SiteMetadata {
+	siteTitle?: string;
+	description?: string;
+	locale?: string;
```
justlevine, line 7: Here and everywhere, if we don't have a reason to treat an explicit undefined differently than an implicit on, we should add `| undefined` to any `?:` defs

packages/next/src/seo/root/site-metadata/validator.ts
```diff
@@ -0,0 +1,44 @@
...
+export const SiteMetadataSchema = z.object( {
+	generalSettings: z
+		.object( {
+			title: z.string().nullable().optional(),
+			description: z.string().nullable().optional(),
+			language: z.string().nullable().optional(),
+		} )
+		.nullable()
+		.optional(),
+} );
```
justlevine, line 18: What's the point of having both an explicit type _and_ a zod schema?

examples/nextjs/starter/src/app/[[...path]]/page.tsx
```diff
@@ -1,5 +1,20 @@
...
 import { EditorBlocksRenderer } from '@snapwp/blocks';
+import { Metadata } from 'next';
+import { headers } from 'next/headers';
+
+export async function generateMetadata(): Promise< Metadata | undefined > {
+	const headerList = await headers();
+	const uri = headerList.get( 'x-current-path' ) || '/';
+
+	if ( uri === '/' ) {
+		return undefined;
```
Comment (justlevine) [inline examples/nextjs/starter/src/app/[[...path]]/page.tsx:12]: Like to the home page doesn't get SEO or am I misunderstanding?

packages/next/src/seo/root/icons-metadata/validator.ts
```diff
@@ -1,44 +1,44 @@
...
-}
+import type { Validator } from '../types';
+import type { FormattedIconData, IconData, IconsMetaData } from './types';
 
-interface FormattedIconData {
-	url: string;
-	sizes: string;
-}
+import { z } from 'zod';
+/**
```
justlevine, line 5: Why do all JS devs _hate_ whitespace?

justlevine (changes requested):
Putting aside the overdue conversation we need to have about [file patterns](https://rtcamp.slack.com/archives/CK2CEH4F6/p1743148212224629), I'm confused by significant parts of this PR.

Might be clearer after the feedback I left is resolved 🤞

packages/next/src/root-layout/index.tsx
```diff
@@ -39,23 +39,7 @@ export async function RootLayout( {
  * @return Merged metadata.
  */
 export async function generateRootMetaData(): Promise< Metadata > {
```
justlevine, line 41: To keep things stackable and pluggable, we should support/pass through the `params` and `searchParams` (ref https://nextjs.org/docs/app/api-reference/functions/generate-metadata#generatemetadata-function)

packages/next/src/seo/route/site-metadata/validator.ts
```diff
@@ -0,0 +1,76 @@
...
+) => {
+	const parsedData = RouteMetadataSchema.safeParse( data );
+	if ( ! parsedData.success || ! parsedData.data.nodeByUri ) {
+		return {};
+	}
+
+	const node = parsedData.data.nodeByUri;
+	const metadata: SiteMetadata = {};
+
+	switch ( node.__typename ) {
```
justlevine, line 23: 1. NEVER use introspection for data, since it's disabled for public sites.
2. As noted on the queries, we should be able to build these non-deterministally, based on the available `node[ ${key} ]`, instead of needing to explicitly support a node.

packages/next/src/seo/route/index.ts
justlevine: I figured out what I dont like:

These `register*` calls are side effects, but there's multiple exportable things in this file not related to those side effects.

We need to separate the concerns somehow.

packages/next/src/seo/route/schema.ts
```diff
@@ -0,0 +1,25 @@
+import { z } from 'zod';
+
+/**
+ * Schema for validating featured image data.
+ */
+const FeaturedImageSchema = z
```
justlevine, line 6: Putting aside why we need zod instead instead of inferring from either the component or the graphql type, this probably doesn't belong here top-level.

packages/query/src/fragments/seo/root/icons-metadata.graphql
```diff
@@ -1,4 +1,4 @@
-query GetGeneralSettings {
+fragment iconMetadataFrag on RootQuery {
```
justlevine, line 1: (If this needs to be separate from SiteMetadataFrag ):

```suggestion
fragment IconMetadataFrag on RootQuery {
```

packages/query/src/fragments/seo/root/site-metadata.graphql
```diff
@@ -0,0 +1,7 @@
+fragment siteMetadataFrag on RootQuery {
```
justlevine, line 1:
```suggestion
fragment SiteMetadataFrag on RootQuery {
```

packages/next/src/seo/root/index.ts
```diff
@@ -0,0 +1,84 @@
...
+	const fragmentDocMap = Object.fromEntries(
+		Object.entries( plugins ).map( ( [ key, plugin ] ) => {
+			return [ key, plugin.fragmentDoc ];
+		} )
+	);
+
+	// Generate Root Query
+	const rootQuery = generateRootQuery( Object.values( fragmentDocMap ) );
+
+	const { data } = await QueryEngine.apolloClient.query( {
```
justlevine, line 62: This feels like it's happening "too late".

Unless batch queries are supported out of the box, all these `RootLayout` queries (icon meta, site meta, global styles) can really just be fragments on a `RootLayoutQuery`.

Ideally our query would be shared across `generateMetadata()` and the component itself, with the DX layer looking something like this (very bad pseudocode):

```tsx

const generateRootQuery(): DocumentNode<inferred type> {
	const rootQueryFragments: Array<string, FragmentType> = getConfig().fragments?.['rootQuery'];

	return graphql(
		${outputFragmentDocuments( rootQueryFragments ) }

		query GetRootQuery {
			${outputFragmentNames( rootQueryFragments )}
		}
	) as DocumentNode;
}

const rootQuery = generateRootQuery().

const { data, errors } = await QueryEngine.getQuery({
	query: generateRootQuery() // used to infer the return type for `data`.
	...otherArgs
});

... do stuff with the data

```

ayushnirwal: That's the end goal for this dynamic query generation pattern. For now, I want to minimize the scope for it to be just SEO things for 3 reasons 
a) This task only relates to doing SEO things. 
b) I want to wait for @SH4LIN to finish https://github.com/rtCamp/headless/issues/402 so that we can retackle the problem but now decoupled from the underlying client (apollo or tanstack)

packages/next/src/seo/route/index.ts
ayushnirwal: This will be refactored into a class. Register calls would be wrapped in a constructor. Plugins will be added through config management in the constructor.

packages/next/src/seo/template/opengraph-parser.ts
```diff
@@ -0,0 +1,43 @@
...
+	OpenGraphMetadataFragFragment
+> = ( data ) => {
+	if ( ! data.connectedNode ) {
+		return {};
+	}
+	const node = data.connectedNode;
+
+	switch ( node.__typename ) {
+		case 'Post':
+		case 'Page':
```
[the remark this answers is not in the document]
ayushnirwal: 1) type inference for __typename does what enums will do already.
2) We don't wanna really use __typename anyway. But it seems to me to be out of the scope of this PR. @justlevine does this concern pop up in any task already? if not this is gonna be a new task for both frontend and backend.

packages/core/src/utils/html.ts
```diff
@@ -0,0 +1,11 @@
+/**
+ * Removes all HTML tags from a given string.
+ *
+ * @param {string} html - The HTML string to be stripped.
+ * @return The cleaned text without HTML tags.
+ */
+export function sanitizeHtml( html: string ): string {
```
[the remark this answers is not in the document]
ayushnirwal: Yes, as the current grapql schema stands we have 2 options to pull content/excerpt for description meta-data. RAW(worpress template syntax) and RENDERED(HTML).

packages/next/src/seo/template/opengraph-parser.ts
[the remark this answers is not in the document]
justlevine: I was under the impression that we _already_ removed all our introspection requirements. Even if not let's not add _any_ new ones, we can always expose a non-introspection `type` field when really necessary.

Here, doesn't matter bc when I'm back from the weekend I'm going to tell you to refactor the whole thing indeterminately (just like the interface), e.g you use a `node.excerpt` for `rawDescription` not because it's a `Post` but because `excerpt` exists so we prioritize it over something broader like `node.description` or `toExcerpt( node.content, charlen)`

examples/nextjs/starter/src/app/layout.tsx
```diff
@@ -11,14 +12,13 @@ export default function Layout( { children }: PropsWithChildren ) {
...
+ * @return Metadata for SEO.
  */
 export async function generateMetadata(): Promise< Metadata > {
-	const rootMetaData = await generateRootMetaData();
+	const metaData = await getSiteMetadata();
 
 	return {
-		...rootMetaData,
+		...metaData,
 	};
```
justlevine, line 23:
```suggestion
	const metadata = await getSiteMetadata();

	return {
		...metadata,
	};
```

packages/query/src/utils/query-generation.ts
```diff
@@ -0,0 +1,43 @@
...
+        ${ Object.values( fragments ).map( ( fragmentDoc ) =>
+			print( fragmentDoc )
+		) }
+        query getRoot{
+            ${ Object.values( fragments ).map(
+				( fragmentDoc ) => `...${ getFragmentName( fragmentDoc ) }`
+			) }
+        }
+    `;
+}
```
justlevine, line 22: Why isn't prettier/eslint formatting this?

packages/query/src/fragments/seo/template/twitter-metadata.graphql
```diff
@@ -0,0 +1,22 @@
...
+		uri
+
+		... on NodeWithTitle {
+			title
+		}
+
+		... on NodeWithFeaturedImage {
+			featuredImage {
+				node {
+					sourceUrl
```
justlevine, line 13:
```suggestion
				node {
					id
					sourceUrl
```

packages/query/src/fragments/seo/template/open-graph-metadata.graphql
```diff
@@ -0,0 +1,34 @@
...
+			title
+		}
+		... on NodeWithExcerpt {
+			excerpt
+		}
+		... on NodeWithFeaturedImage {
+			featuredImage {
+				node {
+					sourceUrl
+					mediaDetails {
```
justlevine, line 15:
```suggestion
					id
					sourceUrl
					mediaDetails {
```

packages/codegen-config/src/index.ts
```diff
@@ -10,6 +10,8 @@ const baseConfig: CodegenConfig = {
 				enumsAsTypes: true,
 				skipTypename: true,
 				useTypeImports: true,
+				useImplementingTypes: true,
```
justlevine, line 13: Is this supported with the `client` preset ? https://the-guild.dev/graphql/codegen/plugins/presets/preset-client

examples/nextjs/starter/package.json
```diff
@@ -4,7 +4,7 @@
 	"private": true,
 	"type": "module",
 	"scripts": {
-		"dev": "concurrently \"next dev\" \"npm run codegen:watch\"",
+		"dev": "concurrently \"NODE_OPTIONS='--inspect' next dev\" \"npm run codegen:watch\"",
```
justlevine, line 7: EILI5?

packages/query/src/fragments/seo/site/icons-metadata.graphql
```diff
@@ -1,7 +1,6 @@
-query GetGeneralSettings {
+fragment IconMetadataFrag on RootQuery {
 	generalSettings {
 		siteIcon {
-			id
 			mediaDetails {
```
justlevine, line 4: Not sure why this was removed.

```suggestion
		siteIcon {
			id
			mediaDetails {
```

packages/next/src/seo/template/opengraph-parser.ts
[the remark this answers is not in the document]
justlevine: This is what a holistic version of the class looks like. From the parser's perspective it doesn't matter what Object Type something is. If the field exists on the schema, then the parser will try to use that.

```
if ( ! data.connectedNode ) {
	return {};
}

const node = data.connectedNode;

const title = node?.title || node?.name || undefined;

// If there's no description, use the first 150 characters of the content
let description = node?.excerpt || node?.description || null;

if ( ! description ) {
	const trimmedContent = node?.content?.substring( 0, 150 );

	if ( trimmedContent ) {
		description = trimmedContent + '...';
	}
}

return {
	openGraph: {
		title,
		description: description ? sanitizeHtml( description ) : undefined,
	},
};
```

packages/query/src/fragments/seo/template/route-metadata.graphql
```diff
@@ -0,0 +1,25 @@
...
+			author {
+				node {
+					name
+				}
+			}
+		}
+		... on NodeWithExcerpt {
+			excerpt
+		}
+		... on Page {
```
justlevine, line 17:
```suggestion
		... on NodeWithContentEditor {
```

packages/query/src/fragments/seo/template/open-graph-metadata.graphql
ayushnirwal: We are not using apollo caching. No need for 'id'

packages/next/src/seo/template/opengraph-parser.ts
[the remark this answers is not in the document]
ayushnirwal: The reliance on `__typename` comes from codgen types. If the node is not type narrowed before accessing data inside it there will be TS errors.

packages/query/src/fragments/seo/template/twitter-metadata.graphql
justlevine: Incorrect. ID is what allows us us to stich multiple queries together into the same cache. The fact that Apollo does this automatically doesn't change the fact that as I've said before _all nodes_ should include their global `id`>

packages/next/src/seo/template/opengraph-parser.ts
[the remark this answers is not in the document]
justlevine: **Never** compromise code patterns for tooling.

I thought we previously disabled fragment masking, and as noted our ideal parsers should be decoupled ( https://github.com/rtCamp/headless/issues/449 ) and parse _any_ data object.

So either way you probably want to be typing the parameters as unknown and type casting/unwrapping any fragmentrefs

examples/nextjs/starter/package.json
ayushnirwal: > Is it needed for this PR?

Nope. I was having a hard time debugging in the terminal. I can remove it.

packages/query/src/fragments/seo/template/twitter-metadata.graphql
ayushnirwal: Before using fetch-policy: no-cache, there apollo was throwing warnings for this (missing id). Since there were no warnings after that, I was under the impression that ID is not needed anymore.

packages/next/src/seo/template/opengraph-parser.ts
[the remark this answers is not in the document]
ayushnirwal: Reliance on `__typename` removed in https://github.com/rtCamp/snapwp/pull/123/commits/dffb4938703c2dda3db67f8e65b136ced366e205 with the help of typecasting.

examples/nextjs/starter/src/app/layout.tsx
ayushnirwal: Huh?

packages/codegen-config/src/index.ts
ayushnirwal: Accidentally added when trying to fix fragment types. Removed

packages/next/src/seo/template/opengraph-parser.ts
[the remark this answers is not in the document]
ayushnirwal: @justlevine I assumed the type generated from the codegen should be more like the type that the node is cast into right now. 
1) Is that assumption wrong?
2) Is there any easy modification for the codegen config that we can do to make the type so?
3) Should we add it to https://github.com/rtCamp/headless/issues/449 ?

packages/query/src/utils/query-generation.ts
justlevine: Nope, mixed tabs/spaces were committed to this file, and are not getting picked up by our local or CI linters.

 I'm assuming it's because it doesnt recognize the `gql` template literally, and _hoping_ it's because you're using Apollo's literal here instead of GraphQL-Codegen's.

image


(PS: I recommend a plugin like https://marketplace.visualstudio.com/items?itemName=nhoizey.gremlins, and to make sure that your VSCode _and_ GH diff settings are set to show whitespace differences )

examples/nextjs/starter/src/app/layout.tsx
justlevine: Its a camelCase nit. The IRL name is `metadata` not `meta[space]data` and this brings the variable in line with `getSiteMetadata` (lowercase `d`)

packages/next/src/seo/template/opengraph-parser.ts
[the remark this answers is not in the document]
justlevine: @ayushnirwal

> Is that assumption wrong?

Yes. https://the-guild.dev/graphql/codegen/plugins/presets/preset-client#fragment-masking 

(A better explanation, but for a different library is https://gql-tada.0no.co/guides/fragment-colocation)

> Is there any easy modification for the codegen config that we can do to make the type so?

Yes at the expense of ts performance and possibly bundle size.

> Should we add it to https://github.com/rtCamp/headless/issues/449 ?

_If there's anything warranted_, but before we bother, PTAL at this syntax (I used template-parser.tsx since it's a but more complex than the latest version of opengraphql-parser):

```tsx
// This is what the user needs to satisfy. We _could_ make it more complicated with certain parsers by using union types.
// I.e. we _need_ either a `title` or a `name` to be explicitly defined.
interface GraphQLNodeData {
	id: string;
	title?: string | null | undefined;
	name?: string | null | undefined;
	excerpt?: string | null | undefined;
	description?: string | null | undefined;
	content?: string | null | undefined;
	author?:
		| {
				node?:
					| {
							name?: string | null | undefined;
					  }
					| null
					| undefined;
		  }
		| null
		| undefined;
}

/**
 * @param {GraphQLNodeData} node Makes more sense to pass the `connectedNode` than the object that contains it.
 */
export const parseRouteSiteMetadata = < T extends GraphQLNodeData >( // That way their GraphQL types warn if they don't have the minimal props. 
	node: T | null | undefined
): Metadata => {
	const title = node?.title || node?.name || undefined;

	let description = node?.excerpt || node?.description || null;

	// If there's no description, use the first 150 characters of the content.
	// Empty strings are considered explictly set.
	if ( description === null || description === undefined ) {
		const trimmedContent = node?.content?.substring( 0, 150 );

		if ( trimmedContent ) {
			description = trimmedContent + '...';
		}
	}

	const authors = [];

	if ( node?.author?.node?.name ) {
		authors.push( {
			name: node.author.node.name!,
		} );
	}

	return {
		title,
		// Sanitize only if description has a truthy value after processing
		description: description ? sanitizeHtml( description ) : undefined,
		authors,
	} satisfies Metadata;
};
```

And then this can be safely consumed with e.g.


```tsx
/**
 * Testing type safety.
 */
const consumeMyParser = ( data: RouteMetadataFragFragment ): Metadata => {
	const node = data.connectedNode;

	if ( ! node ) {
		return {};
	}

	// If there **was** a need to unwrap a type we'd do it here.
	// e.g. const featuredImage = getFragmentData( FeaturedImageFragFragmentDoc, node?.featuredImage ); 

	return parseRouteSiteMetadata( node );
};
```

examples/nextjs/starter/src/app/layout.tsx
ayushnirwal: > (lowercase d)

Could not see that in the review before. Got it.

packages/query/src/utils/query-generation.ts
ayushnirwal: Got it. Fixed it manually for now.

packages/next/src/seo/template/opengraph-parser.ts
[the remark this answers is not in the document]
ayushnirwal: In your example - 
1. Who has defined the `RouteMetadatFragFragment`? The user or the library(snapwp)?
2. Who has defined the `GraphQLNodeData`?
3. Creating a Fragment (RouteMetadatFragFragment) and a type (GraphQLNodeData) seems redundant and a) bad for DX if the user has defined both b) bad for maintenance if the library has defined both.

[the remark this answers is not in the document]
justlevine: 1. GraphQL codegen. It's _either_ ours or overloaded by the user/snapwp plugin API.
2. It's defined by us as part of the parser (as it is in the code snippet), instead of the current `as unknown as { .... }` coersion happening inside the component
3. So that's the beauty I think. We're only defining things once, and it works with _whatever_ gets autogenerated

examples/nextjs/starter/src/app/[[...slug]]/page.tsx
Comment (justlevine) [inline examples/nextjs/starter/src/app/[[...slug]]/page.tsx:1]: `npm run format`

packages/next/src/seo/icons-parser.ts
```diff
@@ -145,3 +159,20 @@ const formatIcons = ( icons: IconData[] ): FormattedIconData[] =>
...
+export interface IconData {
+	sourceUrl: string;
+	height: string;
+	width: string;
+}
+
+export interface FormattedIconData {
+	url: string;
+	sizes: string;
+}
```
justlevine, line 178: move these to the top of the file with the other exported types

packages/next/src/seo/icons-parser.ts
```diff
@@ -98,10 +93,29 @@ export const getIcons = async (): Promise< IconMetaData > => {
...
 		msApplicationTileIcon: findIconBySize( validIcons, '270x270' ),
+	} );
+};
+
+/**
+ *
+ * @param {IconsMetaData} data Meta data in the internal format
+ * @return Meta data consumable by next
+ */
+const reshape = ( data: IconsMetaData ): Metadata => {
```
justlevine, line 108: Can we rename this to `prepareMetadata()` so it's more self documenting?

packages/next/src/seo/index.ts
```diff
@@ -0,0 +1,163 @@
...
+//@todo attach initialization to lifecycle of the app rather than import
+if ( ! Seo.isInitialized ) {
+	try {
+		Seo.initialize();
+	} catch ( error ) {}
+}
+
+// Rexports for simpler interface
+export const getSiteMetadata = Seo.getSiteMetadata;
+export const getTemplateMetadata = Seo.getTemplateMetadata;
```
justlevine, line 163: Unless the `@todo` obviates this, the class should be in a separate file, and `index.ts` should just contain initialization + reexport.

packages/next/src/seo/index.ts
```diff
@@ -0,0 +1,163 @@
...
+	public static registerPlugin(
+		plugin: MetadataPlugin< any,any >,
+	): void {
+		Seo.plugins.push(plugin);
+	}
+
+	/**
+	 * Loads all default plugins.
+	 */
+	private static registerDefaultPlugins(): void {
```
justlevine, line 50: Can we DRY this out with an array + a loop, instead of manually calling registerPlugin a bunch of times?

packages/next/src/seo/types.ts
```diff
@@ -0,0 +1,10 @@
+import type { Metadata } from 'next';
+
+export type MetadataParser< T > = ( data: T ) => Metadata;
+
+export interface MetadataPlugin< TFrag,TData > {
+	fragmentDoc: TFrag;
+	parser: ( data: TData ) => Metadata;
+    type: "root" | "template"
```
justlevine, line 8: Assuming I understand the purpose of this field, I think it should be called `location` or something else that's more self-documenting

justlevine: Honestly I feel the others could use some work too: `fragment` instead of `fragmentDoc`, and idk if `parser` correctly defines what our metadata callback functions do.

packages/next/src/seo/icons-parser.ts
justlevine, line 1: Lets move all these parser files into their own domain. Eg. `./seo/parsers/icons.ts`

packages/next/src/seo/types.ts
justlevine, line 1: Do we want to colocate this with the SEO class instead of in a separate `types` file?

packages/next/src/seo/utils.ts
```diff
@@ -0,0 +1,23 @@
+export interface Node {
+	description?: string | undefined | null;
+	excerpt?: string | undefined | null;
+	content?: string | undefined | null;
+}
+
+export const getNodeDescription = < T extends Node >(
```
justlevine, line 7: It's not the `NodeDescription` (thats the `description` field). Its either the `MetadataDescription` or alternativly go more colloquial with `getDescriptionFromNode()`

packages/query/src/utils/get-fragment-name.ts
justlevine, line 1: this feels like a good candidate to move into `query-generation.ts`

packages/query/src/index.ts
```diff
@@ -1,3 +1,6 @@
 export * from './__generated/graphql';
 export * from './query-engine';
+export * from './utils/query-generation';
+export type { TypedDocumentNode } from '@apollo/client';
```
justlevine, line 4: We _really_ shouldn't be exporting things from 3rd party libraries.

(and when possible we should be getting our types and graphql functions from our codegen folder)

justlevine (changes requested):
Love it 🚀

Primarily cleanup and tests and this should be ready for merge.

packages/query/src/index.ts
ayushnirwal: Codegen does not provide this type. Its supposed to be imported from client implementation. Or from a lib which every client uses '@graphql-typed-document-node/core'

justlevine: https://the-guild.dev/graphql/codegen/plugins/typescript/typed-document-node also the `graphql` import

Im not saying we need to rewrite our codegen rules altogether bur I do want to double check what client preset gives us.

packages/next/src/seo/seo-manager.ts
```diff
@@ -0,0 +1,165 @@
...
+
+import type { Metadata } from 'next';
+
+export type MetadataParser< T > = ( data: T ) => Metadata;
+
+export interface MetadataPlugin< TFrag, TData > {
+	fragment: TFrag;
+	parseMetadata: ( data: TData ) => Metadata;
+	type: 'root' | 'template';
+}
```
justlevine, line 27: https://github.com/rtCamp/snapwp/pull/123#discussion_r2047986915

packages/query/src/index.ts
justlevine: tl;dr this can be removed. the pattern is fine, but we're importing it directly in the one place we need it.

justlevine (approved):
LGTM, usage docs can be written as a follow up PR.

Merging is blocked by #144 (and will likely require squashing some merge conflicts)

justlevine (changes requested):
Needs the merge conflicts deduped / query engine rework before we can review/merge

ayushnirwal:
@SH4LIN Query Engine rework is complete. PTAL to check if I used the interface correctly. Also, please do one round of manual testing to see if the metadata is showing up in the HTML.

justlevine (approved):

PR #139, opened 2025-03-28
feat: Run npm install (silently) inside npx snapwp

ashutoshgautams:
## What
Run `npm install` silently inside `npx snapwp` command to improve user experience

## Why
Currently, users have to wait for CLI setup and then manually run `npm install`, creating two separate waiting periods. This change automates the installation step, reducing friction in the setup process while also suppressing npm warnings that might be confusing to new users.

### Related Issue(s):
- Part of https://github.com/rtCamp/headless/issues/409

## How
- Created a new `runNpmInstall.cjs` file in the `create-app` directory to handle silent npm installation
- Added this step to the main CLI workflow in `snapwp.cjs` after package versions are updated
- Updated success message in `printSuccessMessage.cjs` to remove the manual `npm install` instruction
- Configured the installation process to suppress non-essential npm output while still showing errors

## Testing Instructions
I'll update the testing instructions section of your PR:

## Testing Instructions
1. Go to the CLI directory:
   ```bash
   cd packages/cli
   ```

2. Build the package:
   ```bash
   npm run build
   ```

3. Link the package globally:
   ```bash
   npm link
   ```

4. Create a temporary test directory elsewhere:
   ```bash
   mkdir -p ~/Desktop/test-snapwp
   cd ~/Desktop/test-snapwp
   ```

5. Run the command to test your changes:
   ```bash
   npx snapwp
   ```

6. Follow the setup prompts and verify:
   - The dependencies are automatically installed after setup
   - The success message no longer includes the `npm install` step
   - The created project works directly with `npm run dev`

7. Clean up after testing:
   ```bash
   npm unlink -g snapwp
   ```

## Screenshots
image

## Additional Info
The implementation uses `execSync` with carefully configured options to minimize output while still showing critical errors. This creates a cleaner, more streamlined experience for users.

## Checklist
- [x] I have read the [[Contribution Guidelines](https://github.com/rtCamp/snapwp/blob/develop/.github/CONTRIBUTING.md)](https://github.com/rtCamp/snapwp/blob/develop/.github/CONTRIBUTING.md).
- [x] My code is tested to the best of my abilities.
- [x] My code passes all lints (ESLint, tsc, prettier etc.).
- [x] My code has detailed inline documentation.
- [ ] I have added unit tests to verify the code works as intended.
- [x] I have updated the project documentation as needed.
- [x] I have added a changeset for this PR using `npm run changeset`.


packages/cli/src/create-app/runNpmInstall.cjs
```diff
@@ -0,0 +1,35 @@
...
+const path = require( 'path' );
+
+/**
+ * Runs npm install silently in the project directory.
+ *
+ * @param {string} projectDirPath - Path to the project directory.
+ * @return {Promise<void>}
+ */
+const runNpmInstall = async ( projectDirPath ) => {
+	console.log( 'Installing dependencies... This may take a few moments.' );
```
justlevine, line 11:
```suggestion
	console.log( 'Installing NPM dependencies... This may take a few minutes.' );
```

.changeset/sixty-apples-rescue.md
```diff
@@ -0,0 +1,5 @@
+---
+"snapwp": minor
```
justlevine, line 2: Changesets is not semantic.

```suggestion
"snapwp": patch
```

justlevine (commented):
1. Wondering if we should provide to `--skip-install-dependencies` (or for `--install-deps=false` or whatever)?

2. I don't like the idea of a `Please wait` static message, especially when this takes >1 min easily on new installs. If we cant figure out a way to bubble up the `npm install` status while suppressing the warnings, then should we consider our own progress indicator?
   > If there's a way to selectively suppress npm messages without adding too much effort / tech debt to this, we should consider what "important messages" we should let through.

@ayushnirwal for final decision on either of these.

packages/cli/src/create-app/runNpmInstall.cjs
```diff
@@ -0,0 +1,35 @@
...
+		execSync( 'npm install --silent --no-fund --no-audit', {
+			cwd: projectDirPath,
+			stdio: [ 'ignore', 'pipe', 'pipe' ], // Suppress most output
+			env: { ...process.env, NPM_CONFIG_LOGLEVEL: 'error' }, // Only show errors
+		} );
+
+		console.log( 'Dependencies installed successfully!' );
+	} catch ( error ) {
+		console.warn(
+			'Warning: Could not automatically install dependencies.'
```
SH4LIN, line 24:
```suggestion
			'Warning: Could not install dependencies.'
```

packages/cli/src/create-app/runNpmInstall.cjs
```diff
@@ -0,0 +1,35 @@
...
+ *
+ * @param {string} projectDirPath - Path to the project directory.
+ * @return {Promise<void>}
+ */
+const runNpmInstall = async ( projectDirPath ) => {
+	console.log( 'Installing dependencies... This may take a few moments.' );
+
+	try {
+		// Run npm install with minimal output
+		execSync( 'npm install --silent --no-fund --no-audit', {
```
SH4LIN, line 15: Use `spawnSync`.
```suggestion
const result = spawnSync('npm', ['install', '--silent', '--no-fund', '--no-audit'], {
    cwd: projectDirPath,
    stdio: ['ignore', 'pipe', 'pipe'], // Suppress most output
    env: { ...process.env, NPM_CONFIG_LOGLEVEL: 'error' }, // Only show errors
    encoding: 'utf-8',
});

// Handle errors
if (result.error) {
    console.error('\nInstallation failed:', result.error.message);
} else if (result.stderr) {
    console.error('\nInstallation warnings/errors:', result.stderr.trim());
} else {
    console.log('\nInstallation finished successfully.');
}
```

SH4LIN:
> 1. Wondering if we should provide to `--skip-install-dependencies` (or for `--install-deps=false` or whatever)?
> 2. I don't like the idea of a `Please wait` static message, especially when this takes >1 min easily on new installs. If we cant figure out a way to bubble up the `npm install` status while suppressing the warnings, then should we consider our own progress indicator?

@justlevine I agree with both options. We should provide a way for developers to skip dependency installation if they want. Also, we should at least show a progress bar when installing dependencies to indicate that the command is running, which will improve the user experience.

packages/cli/src/create-app/runNpmInstall.cjs
ashutoshgautams: Oh, of course, `spawnSync` does provide better debugging info and streamed output handling.

packages/cli/src/create-app/runNpmInstall.cjs
```diff
@@ -0,0 +1,51 @@
...
+
+/**
+ * Runs npm install silently in the project directory.
+ *
+ * @param {string} projectDirPath - Path to the project directory.
+ * @return {Promise<void>}
+ */
+const runNpmInstall = async ( projectDirPath ) => {
+	console.log(
+		'Installing NPM dependencies... This may take a few minutes.'
```
ayushnirwal, line 12: - Use [ora](https://github.com/sindresorhus/ora) to add a spinner to the message.
- Change the message to say "Installing NPM dependencies".
- Show the message after spawning is successful.

packages/cli/src/snapwp.cjs
ayushnirwal, line 1: Add a flag "--skip-install". When passed `runNpmInstall` should not run.

ayushnirwal (changes requested):

packages/cli/src/create-app/runNpmInstall.cjs
```diff
@@ -0,0 +1,83 @@
...
+
+/**
+ * Runs npm install in the project directory with improved output handling.
+ *
+ * @param {string} projectDirPath - Path to the project directory.
+ * @param {Object} options - Configuration options.
+ * @param {boolean} options.skipInstall - Whether to skip the installation step.
+ * @return {Promise<boolean>} - Returns true if installation was successful or skipped.
+ */
+const runNpmInstall = async ( projectDirPath, options = {} ) => {
```
ayushnirwal, line 12: Bad Pattern. If `--skip-install` just don't call this function in the main process.

packages/cli/src/snapwp.cjs
```diff
@@ -64,8 +69,13 @@ const DEFAULT_PROJECT_PATH = './snapwp-app';
 		// Step 6: Update package versions
 		await updatePackageVersions( projectDirPath );
 
-		// Step 7: Print instructions
-		printSuccessMessage( projectDirPath, false );
+		// Step 7: Install dependencies (skip if flag is set)
+		const installSuccess = await runNpmInstall( projectDirPath, {
```
ayushnirwal, line 73: Rather than getting an installSuccess as an bool, Let `runNpmInstall` throw. Derive if args for `printSuccessMessage` from that.

ayushnirwal (changes requested):

packages/cli/src/create-app/runNpmInstall.cjs
```diff
@@ -0,0 +1,83 @@
+const { spawnSync } = require( 'child_process' );
+const ora = require( 'ora' ); // Add ora for spinner
```
justlevine, line 2: This should be in the package.json, no?

ayushnirwal: Yup

ayushnirwal (approved):

justlevine (commented):
@ayushnirwal @ashutoshgautams `npx snapwp` is working for you? 

Following the steps in the PR description I got

```console
justl@rt-headless-web:/var/www/html/repos$ npx snapwp
/var/www/html/repos/snapwp/packages/cli/dist/create-app/runNpmInstall.cjs:2
const ora = require( 'ora' );
            ^

Error [ERR_REQUIRE_ESM]: require() of ES Module /var/www/html/repos/snapwp/packages/cli/node_modules/ora/index.js from /var/www/html/repos/snapwp/packages/cli/dist/create-app/runNpmInstall.cjs not supported.
Instead change the require of index.js in /var/www/html/repos/snapwp/packages/cli/dist/create-app/runNpmInstall.cjs to a dynamic import() which is available in all CommonJS modules.
    at Object.<anonymous> (/var/www/html/repos/snapwp/packages/cli/dist/create-app/runNpmInstall.cjs:2:13) {
  code: 'ERR_REQUIRE_ESM'
}

Node.js v20.18.2
```

Switching it to a dynamic import ( `const { default: ora } = await import( 'ora' );` inside the `runNpmInstall()` command fixes that, but the spinner doesnt spin.

_And_ cancelling (`ctrl + c`) causes `✔ Dependencies installed successfully!` to be erroneously displaying.

packages/cli/src/create-app/runNpmInstall.cjs
```diff
@@ -0,0 +1,72 @@
...
+ *
+ * @param {string} projectDirPath - Path to the project directory.
+ * @throws {Error} - Throws an error if installation fails.
+ */
+const runNpmInstall = async ( projectDirPath ) => {
+	// Initialize spinner
+	const spinner = ora( {
+		text: 'Installing NPM dependencies',
+		color: 'cyan',
+	} );
```
justlevine, line 15: This was the change I made to get around the ERR_REQUIRE_ESM, not sure if it's the cause of my other issues.

```suggestion
const { spawnSync } = require( 'child_process' );

/**
 * Runs npm install in the project directory with improved output handling.
 *
 * @param {string} projectDirPath - Path to the project directory.
 * @throws {Error} - Throws an error if installation fails.
 */
const runNpmInstall = async ( projectDirPath ) => {
	// Import ora dynamically
	const { default: ora } = await import( 'ora' );

	// Initialize spinner
	const spinner = ora( {
		text: 'Installing NPM dependencies',
		color: 'cyan',
	} );
```

justlevine (approved):
🚀🚀🚀

image

PR #144, opened 2025-04-03
feat: decouple query engine

SH4LIN:
## What

- This PR removes Apollo Client from the query package and creates a separate package for it, along with the TanStack package. These packages can be used to create an instance of the adapter class, which in turn provides functions to make GraphQL requests. 

This adapter class implements the base `QueryClientAdapter` interface, and its structure is as follows:

```JS
export interface QueryClientAdapter<
	TClient = unknown,
	TClientOptions = unknown,
> {
	init( options?: TClientOptions ): TClient;
	getClient( options?: TClientOptions ): TClient;
	getServerClient( options?: TClientOptions ): TClient;
	useClient( client: TClient | undefined ): TClient | undefined;
	fetchQuery< TData >( {
		key,
		query,
		options,
	}: {
		key: string[];
		query: DocumentNode | TypedDocumentNode< TData >;
		// Keeping this unknown because, at this level, we don't know the query options. If we use a generic type, the class implementing this function will add a guardrail for that particular client, but it will cause a type mismatch error.
		options?: unknown;
	} ): Promise< TData >;
	// eslint-disable-next-line @typescript-eslint/no-explicit-any -- Generic type for query options.
	useQuery< TData >( {
		key,
		query,
		options,
	}: {
		key: string[];
		query: DocumentNode | TypedDocumentNode< TData >;
		// Keeping this unknown because, at this level, we don't know the query options. If we use a generic type, the class implementing this function will add a guardrail for that particular client, but it will cause a type mismatch error.
		options?: unknown;
	} ): TData;
	QueryProvider: React.ComponentType<
		PropsWithChildren< { client: TClient } >
	>;
}
```

## Why

- Apollo client was hardcoded inside the query package which makes it hard to replace or override with other client.

### Related Issue(s):

- Part of: https://github.com/rtCamp/headless/issues/402

## How

- We created an interface that must be implemented by the chosen client adapter class. This adapter class must be a singleton, and its instance should be passed from the SnapWP config.

## Checklist


packages/query/src/query-repository/index.ts
```diff
@@ -0,0 +1,95 @@
...
+} from '@snapwp/core';
+import parseGeneralSettings from '@/utils/parse-general-settings';
+import { getConfig } from '@snapwp/core/config';
+import { QueryAdapterRegistry } from '@/query-adapter-registry';
+import type { BlockData } from '@snapwp/types';
+
+/**
+ * Singleton class to handle GraphQL queries using Apollo.
+ */
+export class QueryRepository {
```
justlevine, line 22: What were you trying to address by renaming this class?

Either way lets revert for now so the diff isnt obscured by the rename.

packages/query/src/query-repository/index.ts
```diff
@@ -0,0 +1,95 @@
...
+	/**
+	 * Fetches global styles.
+	 * @return The template data fetched for the uri.
+	 */
+	static getGlobalStyles = async (): Promise< GlobalHeadProps > => {
+		const data = await QueryAdapterRegistry.adapter.fetchQuery( {
+			key: [ 'globalStyles' ],
+			query: GetGlobalStylesDocument,
+			options: {},
```
justlevine, line 31: Note, while the `Adapter.fetchQuery()` accepts these, we still need a way of changing these in userland.

packages/types/src/query-engine-base/index.ts
```diff
@@ -0,0 +1,37 @@
...
+	getServerClient( options?: TClientOptions ): TClient;
+	useClient( client: TClient | undefined ): TClient | undefined;
+	fetchQuery< TData >( {
+		key,
+		query,
+		options,
+	}: {
+		key: string[];
+		query: DocumentNode | TypedDocumentNode< TData >;
+		// Keeping this unknown because, at this level, we don't know the query options. If we use a generic type, the class implementing this function will add a guardrail for that particular client, but it will cause a type mismatch error.
```
justlevine, line 20: Can you clarify the where/how the type mismatch error occurs?

My gut want's this to be `Partial<TClientOptions>` (or similar),

packages/types/src/query-engine-base/index.ts
justlevine, line 1: Since this lives in types I strongly recommend we annotate this entire Interface to make it easy for implementers to create their own adapters without needing multiple references.

packages/types/src/query-engine-base/index.ts
```diff
@@ -0,0 +1,37 @@
...
+		options?: unknown;
+	} ): TData;
+	QueryProvider: React.ComponentType<
```
justlevine, line 34: NEVER use the `React.` namespace when you can `import type { something } from 'react';`

(TODO: we need to make this an eslint rule)


```suggestion
	QueryProvider: ComponentType<
```

packages/types/src/query-engine-base/index.ts
```diff
@@ -0,0 +1,37 @@
...
+	init( options?: TClientOptions ): TClient;
+	getClient( options?: TClientOptions ): TClient;
+	getServerClient( options?: TClientOptions ): TClient;
+	useClient( client: TClient | undefined ): TClient | undefined;
+	fetchQuery< TData >( {
+		key,
+		query,
+		options,
+	}: {
+		key: string[];
```
justlevine, line 18: What's the `key` used for?

packages/tanstack/src/query-provider/index.tsx
```diff
@@ -0,0 +1,30 @@
+'use client';
+
+import { type JSX, type PropsWithChildren } from 'react';
+import { type QueryClient, QueryClientProvider } from '@tanstack/react-query';
+
+/**
+ * TanStack Query Provider component.
+ * @param root0 - The root object.
+ * @param root0.children - The child components.
+ * @param root0.client - The TanStack Client instance.
```
justlevine, line 10: Conventions in the rest of the codebase has us calling these `props.` not `root#`

packages/apollo/src/query-engine/index.ts
justlevine, line 1: Assumedly the internal file structure should be self-documenting to what it contains, so we probably want the adapters to look closer to

```
.src/
|- adapter/
|  |- index.tsx
|  |_ provider.tsx
|- index.ts (entrypoint)
```

packages/types/src/query-engine-base/index.ts
SH4LIN: In TanStack, the key is used for caching directly within the fetch function, but in Apollo, it isn't used yet.

packages/query/src/query-repository/index.ts
SH4LIN: I'll revert the name for now.

packages/types/src/query-engine-base/index.ts
SH4LIN: When we specify a generic type TQueryOptions, the class implementing this interface will add guardrails for this generic type so that users using a particular adapter aren't able to add random data. However, TypeScript is throwing an error saying that TQueryOptions extends QueryOptions from ApolloAdapter doesn't match TQueryOptions.

image
image
image

examples/nextjs/starter/snapwp.config.ts
```diff
@@ -1,5 +1,8 @@
 import type { SnapWPConfig } from '@snapwp/core/config';
+import { ApolloQueryClientAdapter } from '@snapwp/apollo';
 
-const config: SnapWPConfig = {};
+const config: SnapWPConfig = {
+	queryEngine: ApolloQueryClientAdapter.getInstance(),
```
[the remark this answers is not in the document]
SH4LIN: I thought of the same but With string or new callable() user won't be able to pass QueryConfigs.

examples/nextjs/starter/package.json
```diff
@@ -24,6 +24,7 @@
 		"@snapwp/core": "file:../../../packages/core",
 		"@snapwp/next": "file:../../../packages/next",
 		"@snapwp/query": "file:../../../packages/query",
+		"@snapwp/apollo": "file:../../../packages/apollo",
```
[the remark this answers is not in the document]
ayushnirwal: My suggestion - `@snapwp/plugin-<system>-<name>`

"plugin" describes its a plugin to something in the snapwp system.
"<system>" describes what system is the package a plugin to. In this case, its 'query'.
"<name>" describes what will be added through this plugin. In this case. its 'apollo'. 

'query-adapter' is omitted on purpose, that being the `apollo-query-adapter` says this will add a query adapter for apollo. `apollo` says this will add anything required for apollo integration, which includes the adapter and other things that can be added in the future. (eg: default config for overloading, Hydration boundaries, react hooks, debugging etc.)

examples/nextjs/starter/snapwp.config.ts
```diff
@@ -1,5 +1,8 @@
 import type { SnapWPConfig } from '@snapwp/core/config';
+import { ApolloQueryClientAdapter } from '@snapwp/apollo';
```
ayushnirwal, line 2: I see no reason to hold on to the 'Adapter' semantics here. It can be just name 'ApolloQueryEngine'

ayushnirwal, line 2: I see no reason to hold on to the 'Adapter' semantics here. It can be just named 'ApolloQueryEngine'

packages/query/src/query-engine/index.ts
```diff
@@ -1,105 +1,37 @@
...
-
-				// If there are networkError throw the error with proper message.
-				if ( error.networkError ) {
-					// Throw the error with proper message.
-					throw new Error(
-						getNetworkErrorMessage( error.networkError )
-					);
-				}
-			}
+		const data = await QueryAdapterRegistry.adapter.fetchQuery( {
```
ayushnirwal, line 28:
```suggestion
		const data = await getConfig().queryEngine.fetchQuery( {
```

packages/tanstack/src/adapter/index.ts
```diff
@@ -0,0 +1,178 @@
...
+	init( options?: QueryClientConfig ): QueryClient {
+		return new QueryClient( options );
+	}
+
+	/**
+	 * Returns the TanStack QueryClient instance.
+	 * @param { QueryClientConfig } options - Generic client options (not used in this implementation).
+	 * @return The QueryClient instance.
+	 */
+	// @ts-ignore
```
ayushnirwal, line 68: Add explanation for the ts-ignore

packages/query/src/utils/parse-template.ts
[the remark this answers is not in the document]
SH4LIN: I believe this might be beyond the current scope of work. I'm happy to refactor it in a separate PR.

packages/query/src/query-repository/index.ts
SH4LIN: @justlevine We may have to discuss the ways on how to allow that. I'll add my suggestions. Once all the actionable feedback is completed.

CC: @ayushnirwal

packages/types/src/query-engine-base/index.ts
SH4LIN: @ayushnirwal This base type is getting used in snapwp config manager in core package. And core package only uses types package. Moving this into query package will cause circular dependency with core package

examples/nextjs/starter/package.json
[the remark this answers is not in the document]
SH4LIN: Hi @justlevine, should I go ahead with @ayushnirwal's suggestion, or would you prefer we discuss it further?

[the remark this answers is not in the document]
justlevine: @SH4LIN we need to discuss this further.

@ayushnirwal if I'm understanding the patterns, then applied we'd have:

- `plugin-query-apollo-client` (As noted [elsewhere](https://github.com/rtCamp/snapwp/pull/144#discussion_r2033408684), the library name is called Apollo *Client*)
- `plugin-query-tanstack-query` (Because like apollo, the there's 20 "Tanstack" libraries, some of which we might choose to adopt some-day)
- `plugin-seo-yoast` (or `plugin-seo-yoast-seo` if we want consistency).
- `plugin-ecommerce-woocommerce` (? or is `<system>` package and should be `next` or `blocks` or idk)

( Personally, this feels very unwieldy, and don't really scale well. )

1. Why prefix with `plugin-*`, when it seems the larger ecosystem pattern is a `*-plugin` suffix?
2. How do we handle "plugins" that don't map cleanly to a single package?
3. Is _everything_ optional a `plugin`, or only things that are registered via `SnapWPConfig.plugins: String[]` (or whatever our Extensions API entrypoint is called).
4. Is there anything in particular we should do to handle the confusion between the different meanings of the _word_ "plugin"? We've got:
   - WordPress "plugin"
   - The typical "plugin" for an NPM dependency
   - Whatever it is we're doing

examples/nextjs/starter/snapwp.config.ts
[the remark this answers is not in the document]
justlevine: @SH4LIN can you clarify? I don't think I have enough FE knowledge, but my gut says this is a engineering problem - which means it has an engineering solution that can be evaluated for viability.

For the sake of code splitting and performance, _nothing_ should be initialized directly in `snapwp.config.ts` but only if/when/on the routes that the functionality is needed. If there's something can't be lazy-loaded, then that's probably a sign we need to rearchitect it better (although possibly as a follow-up to this when we get to the Extension API)

packages/types/src/query-engine-base/index.ts
justlevine: So then then it should be move off of this interface and onto the one extended by Tanstack.

packages/query/src/query-repository/index.ts
justlevine: > Initially, this function was handling QueryEngine and all the API calling functions. Now that QueryEngine has been removed from this class, having the QueryEngine class name no longer makes sense. The class only contains API calling functions, so it is now categorized as a repository and its name has been changed to QueryRepository.

My _only_ issue with the above is that this isn't (semantically) a "repository" of queries or the resolved data, it's just an abstraction layer for the underlying query engine.

What do you think about `QueryClient` (my preference, since it's self documenting from a user POV) or `QueryService` (closer to what the internal pattern is vs a Repository)

(Either way still hold off on changing for now)

packages/apollo/src/engine/query-provider.tsx
```diff
@@ -0,0 +1,31 @@
...
+ * @param { Object } props - The root object.
+ * @param { ReactNode | undefined } props.children - The child components.
+ * @param { ApolloClient< NormalizedCacheObject > } props.client - The Apollo Client instance.
+ *
+ * @return The rendered ApolloProvider component.
+ */
+export const ApolloQueryProvider: ( {
+	client,
+	children,
+}: React.PropsWithChildren< {
```
justlevine, line 22: https://github.com/rtCamp/snapwp/pull/144#discussion_r2027611583

packages/types/src/query-engine-base/index.ts
```diff
@@ -0,0 +1,84 @@
+import { type ComponentType, type PropsWithChildren } from 'react';
```
justlevine, line 1:
```suggestion
import type { ComponentType, PropsWithChildren } from 'react';
```

packages/query/src/query-engine/index.ts
```diff
@@ -1,105 +1,36 @@
...
-
-				// If there are networkError throw the error with proper message.
-				if ( error.networkError ) {
-					// Throw the error with proper message.
-					throw new Error(
-						getNetworkErrorMessage( error.networkError )
-					);
-				}
-			}
+		const data = await getConfig().queryEngine.fetchQuery( {
```
justlevine, line 27: Am I mistaken or in react do we always want to be object destructuring `getConfig()` versus direct access?

cc @ayushnirwal per https://github.com/rtCamp/snapwp/pull/144#discussion_r2032485169

ayushnirwal: People generally lean towards destructuring. cc: @SH4LIN

examples/nextjs/starter/snapwp.config.ts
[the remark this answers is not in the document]
SH4LIN: So, if the user passes a string (i.e., the path to their implementation of the adapter), how will they pass the configuration to initialize their engine? And if they pass a callable, how will they pass the engine configuration in that case?

packages/query/src/query-repository/index.ts
SH4LIN: I am of the opinion that we should use QueryService.

packages/types/src/query-engine-base/index.ts
SH4LIN: That would raise a question that should we pass key [here](https://github.com/rtCamp/snapwp/pull/144/files#diff-4dcd4a6579e9efbef0c76f97f7e47dba4c38da45a5c8d6bd446c9d8ce7ddabc5R64) or not.

examples/nextjs/starter/package.json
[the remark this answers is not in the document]
ayushnirwal: > Why prefix with plugin-*, when it seems the larger ecosystem pattern is a *-plugin suffix?

Not true, there are examples for both sides - eslint, prettier, vite do plugin-* and webpack does  *-plugin. My reasoning was to have a name that goes from a bigger scope to a smaller scope.

> How do we handle "plugins" that don't map cleanly to a single package?

Any plugin has to conform to that mapping due to the plugin architecture we are moving towards. Blocks have `BlockManager`, Query has `QueryEngine` etc. `plugin-woocommerce` can hypothetically export both 'plugin-woocommerce/seo' to add meta-data and 'plugin-woocommerce/blocks' to add block renderers.

 > Is everything optional a plugin, or only things that are registered via SnapWPConfig.plugins: String[] (or whatever our Extensions API entrypoint is called).

It should be.

>Is there anything in particular we should do to handle the confusion between the different meanings of the word "plugin"? We've got:
 
What to call a plugin if not a plugin? Prettier and eslint plugins add the prefix `prettier-plugin` and `eslint-plugin`. We _can_ do that.

packages/query/src/utils/parse-template.ts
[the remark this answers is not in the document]
justlevine: https://github.com/rtCamp/headless/issues/449

packages/query/src/query-repository/index.ts
justlevine: Can you share why?

SH4LIN: TanStack's client is of type QueryClient, Which will cause confusion. With GraphQL context QueryClient is something you call fetchQuery function. Our class provides wrapper functions that calls the API, so QueryService is the next suitable option

packages/types/src/query-engine-base/index.ts
ayushnirwal: Any resolution on this @SH4LIN

ayushnirwal (changes requested):

packages/types/src/query-engine-base/index.ts
SH4LIN: @ayushnirwal I am working on it.

packages/apollo/package.json
```diff
@@ -0,0 +1,43 @@
...
+		"dev": "tsc-watch -b --onSuccess \"npm run resolve-paths\"",
+		"resolve-paths": "resolve-tspaths",
+		"lint": "eslint . --config ../../.eslintrc.json",
+		"lint:fix": "eslint . --config ../../.eslintrc.json --fix"
+	},
+	"dependencies": {
+		"@snapwp/core": "*"
+	},
+	"peerDependencies": {
+		"@apollo/client": "^3.13.4"
```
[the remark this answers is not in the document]
SH4LIN: @justlevine Let's assume that if the end user consuming our Apollo adapter package hasn't installed the @apollo/client package, then the peer dependency will install it at the root of node_modules, allowing them to use Apollo Client directly inside examples/nextjs/starter. If they have already installed @apollo/client, we want to constrain the version they can use to ensure compatibility with our package — and using peerDependencies is the way to do that.

packages/types/src/query-engine-base/index.ts
SH4LIN: @ayushnirwal This will have the same issue - https://github.com/rtCamp/snapwp/pull/144#discussion_r2034655518

packages/apollo/package.json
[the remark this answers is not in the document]
justlevine: > then the peer dependency will install it at the root of node_modules

That is **not** how `peerDependencies` work. Users are _supposed_ to install peer dependencies direction in their own `package.json` files. They can opt for the behavior you're describing with the `--legacy-peer-deps` or `--force` flags, but as indicated by the names of those flags this is **not** ideal behavior.

> If they have already installed @apollo/client, we want to constrain the version they can use to ensure compatibility with our package — and using peerDependencies is the way to do that.

This is true whether we use normal deps or peer deps.

---

Rather as I understand it, (@ayushnirwal correct me if I'm wrong) a `peerDependency` is useful when we want to _force_ a user to directly install and manage a dependency.

If a user is installing `@snapwp/apollo-client-plugin` , then it can be assumed they want a compatible version of `@apollo/client` installed, without needing to worry about managing that version directly.

examples/nextjs/starter/package.json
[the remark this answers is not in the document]
justlevine: > What to call a plugin if not a plugin? 

Extensions or addons (?)

packages/apollo/package.json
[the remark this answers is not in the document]
ayushnirwal: > a peerDependency is useful when we want to force a user to directly install.
No, A `peerDependency` is installed. if 
a) user has not mentioned it as their dependency 
b) If the user has mentioned it as their dependency and the versions are not conflicting)
> manage a dependency (version).
Yes, see b). If the versions are conflicting then user will need to resolve it in their project
> If a user is installing @snapwp/apollo-client-plugin , then it can be assumed they want a compatible version of @apollo/client installed, without needing to worry about managing that version directly.
Exactly, this is what `peerDependencies` does. But in the case user has `@apollo/client` as their dependency
`dependencies` will install 2 versions.
`peerDependencies` will ask the user to resolve the version conflict.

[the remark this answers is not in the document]
justlevine: Leaving this for future reference in case this comes up again or I forget we're post-COVID https://github.com/npm/rfcs/blob/main/implemented/0025-install-peer-deps.md

examples/nextjs/starter/snapwp.config.ts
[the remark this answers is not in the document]
justlevine: So again. My general assumption would be (pseudocode) `const resolved[ key ] =  typeof config[ key ] === 'function' ? config[ key ]() : config[ key ];`.

If that doesnt answer your question then I repeat:

> an you clarify? I don't think I have enough FE knowledge, but my gut says this is a engineering problem - which means it has an engineering solution that can be evaluated for viability.

This time please _answer_ instead of deflecting with another question ;-)

justlevine (changes requested):
The remaining open comments neeed a discussion/decision before we can proceed.

examples/nextjs/starter/package.json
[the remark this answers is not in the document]
justlevine: Okay we're simplifying things to just `@snapwp/plugin-{purpose}` or `snapwp-plugin-{purpose}` (for 3rd party).

So here we'll use `plugin-apollo-client` and `plugin-tanstack-query`.

@SH4LIN the remaining comments are all on the main `query` package, so go ahead and rename these, no need to worry about the diff.

packages/types/src/query-engine-base/index.ts
ayushnirwal: The solution I see here is to update `fetchQueryArgs` to be -

```
export type fetchQueryArgs< TData > =
	| {
			key: string[];
			query: DocumentNode | TypedDocumentNode< TData >;
			options?: unknown; // Intentionally kept unknown for implementer-defined types.
	  }
	| {
			query: DocumentNode | TypedDocumentNode< TData >;
			options?: unknown; // Intentionally kept unknown for implementer-defined types.
	  };
```

In the queries in the query engine. We do pass the key. The underlying implementation may decide to use it or not. In TQ, it is required. In Apollo, we can provide logging with the key ( which @SH4LIN is already doing  ) or just now use it. Any future clients will also have that option. 
Translated into a principle -  While fetching, we give the adapter implementation as much info (through args) as possible ( right now it's key, the interfaces may be expanded in the future ). The implementation (ours or external ) may or may not use it.

ayushnirwal: ☝️ Not a change request. Just a discussion @SH4LIN @justlevine

justlevine: I don't see how that solves the problem, but I _do_ see how that creates direct coupling between the base `fetchQuery` function and the Adapter........

examples/nextjs/starter/package.json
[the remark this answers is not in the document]
SH4LIN: Okay, @justlevine I'll refactor it.

packages/types/src/query-engine-base/index.ts
justlevine: FINAL Changes required to close this thread per call w. Ayush + David:

1. `fetchQueryArgs.key: string[]` => `fetchQueryArgs.name : string`
2. Inside the _Tanstack Adapter_ we'll map `name` to `cacheKey`.

SH4LIN: Got it, Will refactor this as well

packages/types/src/query-engine-base/index.ts
```diff
@@ -0,0 +1,94 @@
...
+	 * React component that provides the query client context.
+	 */
+	QueryProvider: ComponentType< PropsWithChildren< { client: TClient } > >;
+}
+
+export type fetchQueryArgs<
+	TData,
+	TQueryOptions extends unknown | undefined,
+> = {
+	name: string[];
```
justlevine, line 86:
```suggestion
	name: string;
```

packages/core/src/config/snapwp-config-manager.ts
```diff
@@ -46,6 +46,16 @@ export interface SnapWPConfig {
 	 * html-react-parser overload options
 	 */
 	parserOptions?: HTMLReactParserOptions;
+	/**
+	 * Query Engine
+	 */
+	// eslint-disable-next-line @typescript-eslint/no-explicit-any -- We don't know the type of the query engine, so we use `any`. unknown will cause type error.
+	query: QueryEngine< TClient, TClientOptions >;
+}
```
justlevine, line 54:
```suggestion
	/**
	 * Query Engine
	 */
	query: QueryEngine< TClient, TClientOptions >;
}
```

packages/plugin-apollo-client/src/engine/index.ts
```diff
@@ -0,0 +1,227 @@
...
+
+export type clientType = ApolloClient< NormalizedCacheObject >;
+export type clientOptionsType = ApolloClientOptions< NormalizedCacheObject >;
+
+/**
+ * An adapter for Apollo Client that implements the BaseQueryClientEngine interface.
+ * This adapter provides methods for obtaining an Apollo Client instance, executing queries,
+ * and using queries as hooks.
+ */
+export class ApolloQueryClientEngine
```
justlevine, line 32: (Or ApolloClientAdapter)

```suggestion
export class ApolloClientEngine
```

packages/plugin-tanstack-query/src/engine/index.ts
```diff
@@ -0,0 +1,163 @@
...
+import { getGraphqlUrl } from '@snapwp/core/config';
+import { TanStackQueryProvider } from './query-provider';
+import type { BaseQueryClientEngine } from '@snapwp/types';
+import type { DocumentNode } from 'graphql';
+
+/**
+ * TanStack Query Client Adapter that implements the BaseQueryClientEngine interface.
+ * This adapter allows you to work with TanStack Query in a generic way.
+ */
+export class TanStackQueryClientEngine
```
justlevine, line 19:
```suggestion
export class TanStackQueryEngine
```

packages/plugin-tanstack-query/src/engine/index.ts
```diff
@@ -0,0 +1,163 @@
...
+			variables?: Record< string, unknown >;
+		} = FetchQueryOptions & {
+			variables?: Record< string, unknown >;
+		},
+	>( {
+		key,
+		query,
+		options,
+	}: {
+		key: string[];
```
justlevine, line 100: I don't like that eslint/tsc didnt pick this up.

```suggestion
		name: string;
```

packages/plugin-tanstack-query/src/engine/index.ts
```diff
@@ -0,0 +1,163 @@
...
+		query: DocumentNode;
+		options?: TQueryOptions;
+	} ): Promise< TData > {
+		const graphqlUrl = getGraphqlUrl();
+		// Here we assume that options may include variables.
+		return this.getServerClient().fetchQuery( {
+			queryKey: key,
```
justlevine, line 107:
```suggestion
			queryKey: [name],
```

packages/plugin-tanstack-query/package.json
```diff
@@ -0,0 +1,45 @@
...
+	},
+	"dependencies": {
+		"@snapwp/core": "*",
+		"graphql-request": "^7.1.2"
+	},
+	"peerDependencies": {
+		"@tanstack/react-query": "^5.69.0",
+		"graphql": "^16.9.0"
+	},
+	"devDependencies": {
```
justlevine, line 39: I still dont understand when something should be a peer vs a regular dependency, but I _do_ know this is inconsistent with what the other adapter is doing with `@apollo-client`

packages/query/src/query-engine/index.ts
```diff
@@ -124,40 +58,17 @@ export class QueryEngine {
...
-				// If there are networkError throw the error with proper message.
-				if ( error.networkError ) {
-					// Throw the error with proper message.
-					throw new Error(
-						getNetworkErrorMessage( error.networkError )
-					);
-				}
-			}
+		const data = await fetchQuery( {
+			name: [ 'generalSettings' ],
```
justlevine, line 62: Names should be the GraphQL name
```suggestion
			name: 'GetGeneralSettings',
```

packages/query/src/query-engine/index.ts
```diff
@@ -169,94 +80,18 @@ export class QueryEngine {
...
-		if ( ! QueryEngine.isClientInitialized ) {
-			QueryEngine.initialize();
-		}
 		const variables = { uri };
 
-		try {
-			const data = await QueryEngine.apolloClient.query( {
-				query: GetCurrentTemplateDocument,
+		const data = await fetchQuery( {
+			name: [ 'templateData', uri ],
```
justlevine, line 86: Using variables for additional cache keys is an adapter-level concern.
```suggestion
			name: 'GetCurrentTemplate'
```

packages/types/src/query-engine-base/index.ts
```diff
@@ -0,0 +1,94 @@
...
+import type { ComponentType, PropsWithChildren } from 'react';
+
+/**
+ * A generic interface for creating adapters to support different GraphQL clients
+ * both on the server and client side.
+ *
+ * @template TClient - The type of the GraphQL client instance.
+ * @template TClientOptions - The type of options used to initialize the client.
+ */
+export interface BaseQueryClientEngine<
```
justlevine, line 12: `Base` is redundant.

```suggestion
export interface QueryClientEngine<
```

package.json
```diff
@@ -20,6 +20,8 @@
 		"build:codegen": "npm run build -w @snapwp/codegen-config",
 		"build:types": "npm run build -w @snapwp/types",
 		"build:snapwp": "npm run build -w snapwp",
+		"build:tanstack": "npm run build -w @snapwp/plugin-tanstack-query",
```
justlevine, line 23: I don't mind dropping the `plugin-` prefix, but we should at least call by the rest of the name e.g. `:tanstack-query`

packages/plugin-tanstack-query/src/engine/query-provider.tsx
```diff
@@ -0,0 +1,30 @@
...
+
+/**
+ * TanStack Query Provider component.
+ * @param { Object } props - The root object.
+ * @param { ReactNode | undefined } props.children - The child components.
+ * @param { QueryClient } props.client - The TanStack Client instance.
+ *
+ * @return The rendered TanStackProvider component.
+ */
+export const TanStackQueryProvider: ( {
```
justlevine, line 14:
```suggestion
export const QueryProvider: ( {
```

packages/plugin-apollo-client/src/engine/query-provider.tsx
```diff
@@ -0,0 +1,31 @@
...
+/**
+ * Apollo Query Provider component.
+ *
+ * @param { Object } props - The root object.
+ * @param { ReactNode | undefined } props.children - The child components.
+ * @param { ApolloClient< NormalizedCacheObject > } props.client - The Apollo Client instance.
+ *
+ * @return The rendered ApolloProvider component.
+ */
+export const ApolloQueryProvider: ( {
```
justlevine, line 19:
```suggestion
export const QueryProvider: ( {
```

justlevine (changes requested):
I believe this cleanup is the last round of changes, and then we should be ready to document and merge 🤞

packages/types/src/query-engine-base/index.ts
SH4LIN: @justlevine Tanstack expects this to be an array of string

packages/plugin-tanstack-query/src/engine/index.ts
SH4LIN: We are fulfilling the tanstack requirement here of having array of string, but what if developer actually wants to pass the array of string over here?

packages/query/src/query-engine/index.ts
SH4LIN: https://github.com/rtCamp/snapwp/pull/144#discussion_r2053498679

Tanstack uses this string as a cache key, so here we are caching each page individually by having uri as part of the cache key

packages/types/src/query-engine-base/index.ts
justlevine: Then _Tanstack_ should wrap the string `name` into an `[ name ]` before it uses it as a `queryKey`. 

And an `Options.additionalCacheKeys: string[]` for tanstack users in userland.

packages/query/src/query-engine/index.ts
justlevine: If (_big_ if) that's a valid implementation concern, then it's a _tanstack_ concern, so _tanstack_ should be responsible for extracting _any_ variables into their relevant cache keys.

packages/types/src/query-engine-base/index.ts
SH4LIN: @justlevine Problem lies in Query package where we don't know which client is getting used so we won't be able to pass array of name from there, right now we need to pass it from there.

packages/query/src/query-engine/index.ts
SH4LIN: It's not about query variable, tanstack simply uses array of key as a cache key, so We want individual cache for each page then we pass the uri which will be dynamic at the same time allows to create unique key for each page so they can be cached properly.

packages/types/src/query-engine-base/index.ts
ayushnirwal: @SH4LIN I belive this was resolved in the huddle?

SH4LIN: Yes.

SH4LIN:
@justlevine This PR is ready for re-review

packages/plugin-tanstack-query/src/engine/index.ts
justlevine: Is nobody else concerned that our tooling didnt notice this? @sh4lin @ayushnirwal

package.json
SH4LIN: Then will it be @snapwp/tanstack-query?

justlevine: I'm referring to the script names needing to be updated to reflect the package names: `npm run build:tanstack` => `npm run build:tanstack-query` etc,

packages/plugin-tanstack-query/package.json
```diff
@@ -0,0 +1,45 @@
+{
+	"name": "@snapwp/plugin-tanstack-query",
+	"version": "0.1.0",
+	"license": "AGPL-3.0",
+	"description": "TanStack Client integration for SnapWP.",
```
justlevine, line 5:
```suggestion
	"description": "TanStack Query integration for SnapWP.",
```

justlevine (changes requested):
Plus the two preexisting open conversations.

examples/nextjs/starter/snapwp.config.ts
```diff
@@ -1,5 +1,16 @@
...
+	type clientOptionsType,
+} from '@snapwp/plugin-apollo-client';
 
-const config: SnapWPConfig = {};
+const config: SnapWPConfig< clientType, clientOptionsType > = {
+	query: {
+		// TODO: Find a way to fix this type error
+		// @ts-expect-error TS2419: Class 'ApolloClientEngine' is not generic as the base interface so typescript is throwing an error.
+		engine: ApolloClientEngine,
+	},
```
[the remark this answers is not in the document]
ayushnirwal: https://github.com/rtCamp/snapwp/pull/144#issuecomment-2824030156 May solve this.

packages/plugin-tanstack-query/package.json
justlevine: Bump @sh4lin @ayushnirwal

justlevine (changes requested):
Still two open conversations. In the meantime we can start working on docs in parallel.

Edit: Should also be safe to add a changeset now. Anything _breaking_ is a `minor`, anything nonbreaking is a `patch`. Change to the starter should bump `snapwp` .

packages/plugin-tanstack-query/package.json
SH4LIN: @justlevine I have moved @apollo/client in snapwp/plugin-apollo-client to peerDep.

packages/plugin-tanstack-query/src/engine/index.ts
SH4LIN: @justlevine The time when this happened TS was ignored for the fetchquery for some other reasons and because of that our tooling didn't notice this.

packages/plugin-tanstack-query/package.json
justlevine: But is that where they _should_ be? 

@ayushnirwal mark this thread as resolved so I know this is the correct placement.

SH4LIN: @justlevine We had conversation about this in this thread - https://github.com/rtCamp/snapwp/pull/144#discussion_r2041607457

SH4LIN: Okay so as per my opinion this should be in peerDep. @ayushnirwal Will you be able to validate this?

ayushnirwal: They are fine as per dependencies.

packages/plugin-tanstack-query/package.json
```diff
@@ -0,0 +1,45 @@
...
+		"force-build": "tsc --build --force",
+		"postforce-build": "npm run resolve-paths",
+		"dev": "tsc-watch -b --onSuccess \"npm run resolve-paths\"",
+		"resolve-paths": "resolve-tspaths",
+		"lint": "eslint . --config ../../.eslintrc.json",
+		"lint:fix": "eslint . --config ../../.eslintrc.json --fix"
+	},
+	"dependencies": {
+		"@snapwp/core": "*",
+		"graphql-request": "^7.1.2"
```
ayushnirwal, line 33: graphql-request and @snapwp/core are also peer dependencies.

packages/plugin-apollo-client/package.json
```diff
@@ -0,0 +1,43 @@
...
+		"dev": "tsc-watch -b --onSuccess \"npm run resolve-paths\"",
+		"resolve-paths": "resolve-tspaths",
+		"lint": "eslint . --config ../../.eslintrc.json",
+		"lint:fix": "eslint . --config ../../.eslintrc.json --fix"
+	},
+	"peerDependencies": {
+		"@apollo/client": "^3.13.4"
+	},
+	"dependencies": {
+		"@snapwp/core": "*"
```
ayushnirwal, line 35: @snapwp/core is also a peer dependency.

docs/how-to-overload-query-engine.md
ayushnirwal, line 1: This draft of the doc has too much code and should be minimized by using pseudocode and links to types to be implemented.
Please revert the changes, adding the docs, and do them in a different PR.

ayushnirwal (changes requested):
Few minor things.

docs/how-to-overload-query-engine.md
justlevine: (ref https://github.com/rtCamp/headless/issues/462 )

justlevine (approved):

justlevine (changes requested):
Changing this so I can be pinged when tests are passing

justlevine (approved):

justlevine (changes requested):
@SH4LIN add a changeset and we can merge this

Issue #145, opened 2025-04-03
Ensure 404-ed routes return the correct HTTP Status codes


justlevine:
### What problem does this address?

Currently, when `templateByUri{}` resolves to a _non-existent_ WordPress route, the 404 template gets returned, but since theres a template to return the status code presents as a `Success`, instead of properly indicating (to SEO, crawlers, userland handlers) that there is no WP-route at that page. 

![Image](https://github.com/user-attachments/assets/b06894df-f250-467b-a0f3-2dd6048c8c9e)

### What is your proposed solution?

1. Check the results of `templateByUri` to correctly set the status if a route 404s on WP
   - It might be enough to check `templateByUri?.connectedNode === null` or we might need to expose + test an `is404` or a `responseStatusCode` (if we need to handle `isRestricted` `isUnauthorized`, `isRedirected` states as well)

2. Make the response status code "handleable" in userland.
   - User should be able to redirect, overload the status code, etc without the need to requery themselves. 

### Additional Context

Reported by @kidunot89 via DM - plz lmk if there's anything I missed

MoscoviumAlchemist:
It would make a lot of sense to expose an is404, and also a lot of other WP conditionals would be super useful for similar reasons. Otherwise, how do you know to render the page differently if it's a search vs an archive, vs a tax, etc. I think a ton of edge cases like the 404 will be encountered within this new paradigm, and there's already conventions in place for handling them if they are exposed to the frontend.

justlevine:
Thanks for the feedback @MoscoviumAlchemist 🙏

> also a lot of other WP conditionals would be super useful for similar reasons. 

PS, the `RenderedTemplate` already exposes `{ isComment, isContentNode, isFrontPage, isPostsPage, isTermNode }` alongside the `.connectedNode: UniformResourceIdentifiable` which brings along all the `nodeByUri` conditionals from WPGraphQL core (along with their lifecycle quirks. Part of the work here will be 'stabilizing them' into the `templateByUri` lifecycle ).

Between those and the noted mechanism to modify the GQL fragments ( so users can target their own/third party WPGraphQL fields) and make the response userland accessible, we should be in a good place to handle any edge cases that could come up 🤞

(I still haven't had time to move the GH Issues out of the old internal repo, but if you want to lurk our _very early work_ on that last part, take a look at #123 )

MoscoviumAlchemist:
Thanks for forwarding up with that additional info, I hadn't seen those conditionals exposed previously!

[...]

PR #150, opened 2025-04-09
fix: correctly use named exports instead of `default`

Ta5r:
## What
This PR fixes a syntax/runtime error caused by incorrect usage of :
- the `withSnapWP` import in the `next.config.mjs` file 
- and the import of `DefaultError`.
in the starter.

## Why
The previous import assumed a default export which resulted in:

```bash
> Build error occurred
SyntaxError: The requested module '@snapwp/next/withSnapWP' does not provide an export named 'default'
    at <unknown> (next.config.mjs:1)
"Build failed"
```
and,

```js
import withSnapWP from '@snapwp/next/withSnapWP';
       ^^^^^^^^^^
SyntaxError: The requested module '@snapwp/next/withSnapWP' does not provide an export named 'default'
    at ModuleJob._instantiate (node:internal/modules/esm/module_job:134:21)
    at async ModuleJob.run (node:internal/modules/esm/module_job:217:5)
```

### Related Issue(s):
- Regression in https://github.com/rtCamp/snapwp/pull/147

## How
- Replaces the default import with the correct named import like, for example:
```diff
- import withSnapWP from '@snapwp/next/withSnapWP';
+ import { withSnapWP } from '@snapwp/next/withSnapWP';
```

### Current errors and failures :
- Screenshot 2025-04-09 at 3 32 20 PM
- Screenshot 2025-04-09 at 3 20 35 PM

### After this PR applies :
- Screenshot 2025-04-09 at 3 20 22 PM

## Checklist
-   [x] I have read the [Contribution Guidelines](https://github.com/rtCamp/snapwp/blob/develop/.github/CONTRIBUTING.md).
-   [x] My code is tested to the best of my abilities.
-   [x] My code passes all lints (ESLint, tsc, prettier etc.).
-   [ ] My code has detailed inline documentation.
-   [ ] I have added unit tests to verify the code works as intended.
-   [ ] I have updated the project documentation as needed.
-   [x] I have added a changeset for this PR using `npm run changeset`.

PR #151, opened 2025-04-09
fix: Add `is404` flag for 404ed routes

Swanand01:
## What
This PR adds the `is404` flag for 404ed routes

## Why
Currently, when templateByUri{} resolves to a non-existent WordPress route, the 404 template gets returned, but since theres a template to return, the status code presents as a Success, instead of properly indicating (to SEO, crawlers, userland handlers) that there is no WP-route at that page.

### Related Issue(s):

- https://github.com/rtCamp/headless/issues/451
- https://github.com/rtCamp/snapwp/issues/145

## How
`parseQueryResult` in `packages/query/src/utils/parse-template.ts` is refactored to also parse out the flag `is404`.

## Checklist


packages/query/src/utils/parse-template.ts
```diff
@@ -51,13 +52,15 @@ export function parseQueryResult(
 	}
 
 	const templateByUri = queryData.data?.templateByUri;
+	const is404 = templateByUri?.connectedNode === null;
```
justlevine, line 55:
```suggestion
	// @todo templateByUri should expose a `is404` field to avoid the perf hit from connectedNode.
	const is404 = templateByUri?.connectedNode === null;
```

packages/query/src/utils/parse-template.ts
```diff
@@ -29,6 +29,7 @@ export function parseQueryResult(
 	scripts: EnqueuedScriptProps[] | undefined;
 	scriptModules: ScriptModuleProps[] | undefined;
 	bodyClasses: string[] | undefined;
+	is404: boolean;
```
justlevine, line 32: @ayushnirwal taking into consideration future-compat with other protentional route-based conditionals: is this the ideal schema shape for consumption?

justlevine (changes requested):
LGTM with some small notes below.

That said, since this PR doesn't seem to actually do anything with / allow usage of `is404` _and_ since `connectedNode` is an unnecessary perf hit. I'm going to put this on hold.

If exposing template date or fixing the error codes gets unblocked before `snapwp-helper` can expose a `templateByUri.is404`, then we'll merge this as a temporary fix. Otherwise, we'll update this to use that eventual field.

Swanand01:
@ayushnirwal This PR for now just returns the `is404` property from the `parseQueryResult`. However, it is still not accessible in the user land. Any ideas on how that can be achieved?

Swanand01:
```ts
const { editorBlocks, bodyClasses, stylesheets, scripts, scriptModules, is404 } =
		await getTemplateData( pathname || '/' );
<main>
    <div className="wp-site-blocks">
        { children( { editorBlocks, is404 } ) }
    </div>
</main>
```
```ts
export default function Page() {
	return (
		<TemplateRenderer>
			{ ( { editorBlocks, is404 } ) => {
				return <EditorBlocksRenderer editorBlocks={ editorBlocks } />;
			} }
		</TemplateRenderer>
	);
}
```
We can expose the is404 to user land this way.

justlevine:
> We can redirect to next's 404 page if is404 is true. If next provides ways to attach status code we can do that.

The acceptance criteria for https://github.com/rtCamp/headless/issues/451 is to use WordPress as the source of truth for the 404 template while providing user control of _status codes_ based on WPGraphQL-exposed data.  I don't think a component-level redirect is an "ideal" DX solve for this, but we can refine it in a follow-up PR.

Swanand01:
> We can redirect to next's 404 page if is404 is true.

Not a big fan of doing this, because the user won't be able to control what is to be done if the page is a 404 page.
With my current implementation, they can at least write their own 404 logic in page.tsx.

Swanand01:
@ayushnirwal @justlevine I have updated the current path middleware to perform the is404 check. 
The obvious caveat here is that `QueryEngine.getTemplateData` is called twice for every request, which is why I think that we should have a separate, performant function to check the 404 status for a path.

https://github.com/Swanand01/snapwp/blob/8202409e654cbbb7c394b5c371bfeb91ba4298db/packages/next/src/middleware/current-path.ts#L18-L62

packages/next/src/middleware/current-path.ts
```diff
@@ -1,22 +1,60 @@
...
+			);
+			is404 = fetchedIs404;
+		} catch ( error ) {}
+
 		const response = await next( request, _next );
 
 		if ( response ) {
-			response.headers.set( 'x-current-path', request.nextUrl.pathname );
+			response.headers.set( 'x-current-path', pathname );
+			response.headers.set( 'x-snapwp-is404', String( is404 ) );
```
justlevine, line 50: Why do we need a custom header? Why would we _want_ a custom header?

packages/next/src/template-renderer/index.tsx
```diff
@@ -9,7 +9,10 @@ import type { ReactNode } from 'react';
 
 export type TemplateRendererProps = {
 	getTemplateData?: ( typeof QueryEngine )[ 'getTemplateData' ];
-	children: ( editorBlocks: BlockData[] ) => ReactNode;
+	children: ( args: {
+		editorBlocks: BlockData[];
+		is404: boolean;
```
justlevine, line 14: If we can get this out of `getTemplateData()` we do we also need to expose it directly?

packages/next/src/middleware/current-path.ts
Swanand01: This was added for testing purposes. Removing it now.

packages/next/src/template-renderer/index.tsx
Swanand01: I am not entirely sure what you mean here. I assume you mean one of these:

Why is 404 status added to the response when we can get is404 out of `getTemplateData()`?
- Acceptance criteria. We can get rid of this flag in template renderer if @ayushnirwal says so.

Why do we need a separate function to get 404 status?
- `getTemplateData` processes a big `GetCurrentTemplate` query, which has other fields. All we need for now is the `connectedNode`, so it makes sense to have it separate.

justlevine: Sorry, it was your first guess: I was asking about the line I commented on: the `is404` param that we're passing to the children.

Can you point me to where that acceptance criteria is listed? Afaik I'm the one who specced both #150 and https://gihtub.com/rtCamp/headless/issues/451 not @ayushnirwal (though obviously if Ayush - or you - has a compelling case for why we need it, I'm happy to defer).

Swanand01: From https://github.com/rtCamp/snapwp/issues/145#issuecomment-2788390101:
> This is meant to be a quick fix to add a status code to the response. The changes to the parser may be moved to Template rendering. After https://github.com/rtCamp/snapwp/pull/123

From https://github.com/rtCamp/snapwp/pull/151#issuecomment-2791810084:
> If next provides ways to attach status code we can do that.

justlevine:
> The obvious caveat here is that `QueryEngine.getTemplateData` is called twice for every request,

The bigger caveat is that you shouldn't be querying data in middleware altogether... If you can avoid it.

Not entirely sure the purpose of all this logic, but my assumption is that the status handling needs to happen in a route handler, not middleware. @ayushnirwal ?

Swanand01:
> > The obvious caveat here is that `QueryEngine.getTemplateData` is called twice for every request,
> 
> The bigger caveat is that you shouldn't be querying data in middleware altogether... If you can avoid it.
> 
> Not entirely sure the purpose of all this logic, but my assumption is that the status handling needs to happen in a route handler, not middleware. @ayushnirwal ?

We cannot render JSX (`<TemplateRenderer>`) in a route handler, and there cannot be a route.js file at the same route as page.js. [Ref](https://nextjs.org/docs/app/building-your-application/routing/route-handlers#route-resolution). I'm not sure what can be done here. @ayushnirwal WDYT?

packages/query/src/queries/get-current-template.graphql
```diff
@@ -126,5 +126,8 @@ query GetCurrentTemplate($uri: String!) {
 				...CoreVideoFrag
 			}
 		}
+		connectedNode {
+			id
+		}
```
justlevine, line 131: `templateByUri.is404` is now available on `snapwp-helper:develop` ([ref](https://github.com/rtCamp/snapwp-helper/pull/107)), so go ahead and update this PR to use it instead of `connectedNode`.

Swanand01:
> @SH4LIN I don't know how you arrived at the middleware solution.
> 
> Try using the [notFound](https://nextjs.org/docs/app/api-reference/functions/not-found) function along with [not-found.js](https://nextjs.org/docs/app/api-reference/file-conventions/not-found) file.
> 
> 1. Create a not-found file that fetches template data for WP 404 page. Since this fetch does not rely on any dynamic parameters, it will be statically built.
> 2. In Template Renderer, call the function `notFound`.
> 
> @justlevine Is there a reliable URI to get the 404 template in (1) through templateByURI? or any other way?
> 
> If no, we can use a garbage URI to fetch 404 template data for now and later update the backend to remove the hack.

What if we use `TemplateRenderer` in not-found.tsx?
```tsx
export default function NotFound() {
	return (
		<TemplateRenderer>
			{ ( { editorBlocks } ) => {
				return <EditorBlocksRenderer editorBlocks={ editorBlocks } />;
			} }
		</TemplateRenderer>
	);
}
```

ayushnirwal:
> What if we use TemplateRenderer in not-found.tsx?

Yes...

justlevine:
> Is there a reliable URI to get the 404 template in (1) through templateByURI? or any other way?

There's currently no way to fetch _just_ a template and not a "resolved" URI. Ideally it would be the same results of the previous query - either via the app or because we have server-side caching. I know most things are locked down, but even just requerying the same URI will give us what we need.

> If no, we can use a garbage URI to fetch 404 template data for now and later update the backend to remove the hack.

I'm not opposed to this as a temporary workaround.

packages/query/src/queries/get-current-template.graphql
```diff
@@ -1,5 +1,6 @@
 query GetCurrentTemplate($uri: String!) {
 	templateByUri(uri: $uri) {
+		is404
```
justlevine, line 3: Sort this AZ

examples/nextjs/starter/src/app/[[...path]]/page.tsx
```diff
@@ -4,7 +4,7 @@ import { EditorBlocksRenderer } from '@snapwp/blocks';
 export default function Page() {
 	return (
 		<TemplateRenderer>
-			{ ( editorBlocks ) => {
+			{ ( { editorBlocks, is404 } ) => {
```
Comment (justlevine) [inline examples/nextjs/starter/src/app/[[...path]]/page.tsx:7]:
```suggestion
			{ ( { editorBlocks } ) => {
```

packages/next/src/template-renderer/index.tsx
justlevine: This comment is still valid after our re-scope. Instead of passing `is404` as a prop to the Template Renderer (which it has 0 to do with editor blocks), they should just call `const { is404 } = await getTemplateData()` inside `page.tsx` (and that's what `status-codes.md` should show )

.changeset/tame-beans-pay.md
```diff
@@ -0,0 +1,5 @@
+---
+"@snapwp/query": patch
+---
+
+fix: Add is404 flag for 404ed routes
```
justlevine, line 5: Our new spec we're no longer "fixing" anything, but exposing a prop for developers to consume.

```suggestion
dev: Expose `QueryEngine.getTemplateData().is404` for custom HTTP status code handling.
```

justlevine (changes requested):
Remaining open comments + docs, and this should be good to merge.

(If we cant import `getTemplateData()` to userland until #144 , then we can doc the ideal dx and then put this `on hold` until that's possible. )

justlevine (approved):

PR #153, opened 2025-04-10
update(ci): use snapwp cli to scaffold next application

L0RD-ZER0:
## What
Updates CI to use SnapWP CLI for scaffolding next application instead of using the examples dir directly.

## How
Use `npx snapwp` after installing the local cli package to scaffold the next application in the ci.

## Additional Info
- After the application is scaffolded, the dependencies are updated to use the local repository packages.

## Checklist


mrrobot47 (approved):

.github/workflows/pull_request_examples.yml
```diff
@@ -97,19 +99,24 @@ jobs:
...
+                  cp '${{ github.workspace }}/examples/${{ matrix.example }}/.env.example' .env
                   sed -i 's|NEXT_PUBLIC_WP_HOME_URL=.*|NEXT_PUBLIC_WP_HOME_URL=http://localhost|' .env
+                  npm i '${{ github.workspace }}/packages/cli' --no-save
+                  npx snapwp <<< '.'
+                  npm i
+                  for i in '${{ github.workspace }}/packages/'*; do
+                      if [ -d "$i" ]; then
+                          npm i "$i"  # Swap snapwp dependencies for local packages
+                      fi
+                  done
```
justlevine, line 116: @ayushnirwal / @SH4LIN if there's an _easy_ cleaner way to install local copies of the packages before I merge this spaghetti, please LMK.

Ref: https://rtcamp.slack.com/archives/CK2CEH4F6/p1744296854287099?thread_ts=1744273250.043439&cid=CK2CEH4F6

.github/workflows/pull_request_examples.yml
```diff
@@ -97,19 +99,24 @@ jobs:
...
+                  mkdir -p snapwp-test-app
 
-            - name: Install dependencies
-              working-directory: ${{ github.workspace }}/examples/${{ matrix.example }}
-              run: npm install
-
-            - name: Copy .env.example to .env
-              working-directory: ${{ github.workspace }}/examples/${{ matrix.example }}
+            - name: Scaffold Snapwp App
+              working-directory: ${{ github.workspace }}/snapwp-test-app
```
ayushnirwal, line 105: @L0RD-ZER0 This is what I do to scaffold a project to test the command and packages pre publish

```
npm i
npm run build
npm link snapwp
npm run publish:local
snapwp --proxy
```

If a package is linked snapwp is just an command. no need for npx
npm run publish:local publishes to a local registry (verdaccio)
snapwp --proxy There is a (undocumented) flag which tells snapwp to use packages from the local registry.

L0RD-ZER0: The current CI implementation works by utilizing the template from the local repository and installing the packages in the local repository itself to test them for the applied changes. The resulting workspace created for the CI process closely mirrors the state achieved when using a custom registry with the `--proxy flag`, but without the additional setup required for the custom registry. 

Please review this and confirm if setting up a custom registry is necessary, as there is almost no difference between the two approaches.

.github/workflows/pull_request_examples.yml
```diff
@@ -97,19 +99,24 @@ jobs:
...
               run: |
-                  cp .env.example .env
+                  cp '${{ github.workspace }}/examples/${{ matrix.example }}/.env.example' .env
                   sed -i 's|NEXT_PUBLIC_WP_HOME_URL=.*|NEXT_PUBLIC_WP_HOME_URL=http://localhost|' .env
+                  npm i '${{ github.workspace }}/packages/cli' --no-save
+                  npx snapwp <<< '.'
+                  npm i
+                  for i in '${{ github.workspace }}/packages/'*; do
```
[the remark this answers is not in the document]
L0RD-ZER0: This is inserted by the yaml templating engine used by GitHub actions. It does not need shell expansion. Use of single quotes is intentional to eliminate the factor of glob expansions in the path in the first place.

.github/workflows/pull_request_examples.yml
justlevine: As noted in [Slack](https://rtcamp.slack.com/archives/CK2CEH4F6/p1744273250043439) the goal of this PR (and https://github.com/rtCamp/headless/issues/399) is two-fold:

1. to prevent issues like https://github.com/rtCamp/headless/issues/453 
2. to ensure our CI is testing as much of our IRL workflow as possible.

Both of these are accomplished by updating the CI to follow the local scaffolding steps that @ayushnirwal outlined below.

With that said:

> The resulting workspace created for the CI process closely mirrors the state achieved when using a custom registry with the --proxy flag, but without the additional setup required for the custom registry.

If your argument here is that we can replace the _entire_ use of Verdaccio with a bash script that's just a for-each loop (and @ayushnirwal agrees), then personally I can get behind that because it's one more piece of tech debt that we don't need to test.

If however the CI _only_ "closely mirrors" the local situation, but we still need Verdaccio's `publish:local` and `--proxy` flag IRL, then there would need to be a very strong case made as to why we don't want to be testing our real-world workflow.  

> as there is almost no difference between the two approaches.

If this refers to the effort, then go ahead and change over as described. Otherwise, review any perceived "downsides"/concerns with @mrrobot47 and share back so @ayushnirwal and myself can make an informed decision.

ayushnirwal: @L0RD-ZER0 
The downsides I see are - 
- Missing support for other package managers (yarn).
- Missing support for protected access to test packages ( we may plan to keep some libs closed source )

justlevine: Per today's sync we're keeping verdaccio and are updating the CI accordingly.

mrrobot47 (approved):

justlevine (changes requested):
PR is working (I did a bit of cleanup), but as a result, eslint is failing due to a conflict between our dependencies and next/core-web-vitals.

We need to fix that before we can merge this PR.

(I tried and failed a few approaches and then ran out of time).

.github/workflows/pull_request_examples.yml
```diff
@@ -8,35 +8,38 @@ on:
...
-    NODE_VERSIONS: '[20]'
+    NODE_VERSIONS: '[20, 22]'
     PATHS: |
         nextjs/starter:
           - .github/workflows/pull_request_examples.yml
           - examples/nextjs/starter/**
+          - packages/**
     GRAPHQL_SCHEMA_FILE: '${{ github.workspace }}/schema.graphql'
     SNAPWP_HELPER_REF: 'develop' # TODO: Get from PR description.
+    TEST_APP_DIR: '${{ runner.temp }}/snapwp-test-app' # Use runner.temp to create app outside the main workspace
```
justlevine, line 42: I did not know this was possible. I love it!

@mrrobot47 any concerns we need to be worried about doing it like this?

justlevine: Confirmed via slack : https://rtcamp.slack.com/archives/CK2CEH4F6/p1744969143237679?thread_ts=1744273250.043439&cid=CK2CEH4F6

justlevine (approved):

PR #155, opened 2025-04-14
feat : adds `no-require-imports` lint rule.

Ta5r:
## Checklist

### DIFF
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -36,6 +36,9 @@ module.exports = {
 			'classnames',
 			'lodash',
 		],
+
+		'@typescript-eslint/no-require-imports': 'error',
+
 		'no-restricted-imports': [
 			'error',
 			{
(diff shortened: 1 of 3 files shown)


.eslintrc.cjs
```diff
@@ -172,6 +175,17 @@ module.exports = {
 			},
 		},
 
+		{
+			files: [
+				'**/*.config.{js,ts,cjs,mjs}',
+				'**/config/**/*.{js,ts,cjs,mjs}',
+				'**/.eslintrc.{js,ts,cjs,mjs}',
```
justlevine, line 182: 1. First rule is too loose. _Any_ file that has the word `.config.` in it? 
2. Anything with `.mjs` or `.ts` supports `import` and should be enforced.
3. If you're matching specific files (like a local eslint file) there's no need to regex a bunch of different file etcs.
4. You anyway might consider the `allow` before using the `overrides` https://typescript-eslint.io/rules/no-require-imports/#allow

packages/cli/build.cjs
justlevine, line 1: Unless we're changing this to `.mjs` this can probably be reverted and allowlisted

justlevine (approved):
Take a look at the latest commit - I cleaned it up some.

PR #164, opened 2025-04-22
fix: Images not opening in lightbox

Swanand01:
## What
This PR adds the Lightbox overlay to enable opening images in a lightbox.

## Why
To fix the bug where image would not open in a lightbox.

### Related Issue(s):
Fixes [#328](https://github.com/rtCamp/headless/issues/328)

## How
A `LightboxOverlay` component is added and rendered in the `TemplateRenderer`

## Testing Instructions
1. Add an image to a page using the `core/image` block.
2. Enable the "Enlarge on click" option from the toolbar
3. Open this page in headless frontend
4. Click on the image
5. The image should open in a lightbox.

## Screenshots
image
image

## Checklist


packages/next/src/components/lightbox-overlay.tsx
```diff
@@ -0,0 +1,80 @@
+/**
+ *
+ */
+export function LightboxOverlay() {
+	const closeButtonLabel = 'Close';
+	const closeButtonColor = '#000';
+	const backgroundColor = '#fff';
```
Swanand01, line 7: @ayushnirwal @SH4LIN How do we get these values from the server?
Do I need to call `getGlobalStyles` and then parse the `globalStylesheet`?

SH4LIN: @Swanand01 Where does WP is getting these values from?

Swanand01: https://github.com/WordPress/gutenberg/blob/accde16096464ac832fbe8e249a3b61535800800/packages/block-library/src/image/index.php#L259-L269

Swanand01: These are obtained from the global styles if theme.json is present.

Swanand01: Using css vars for now.

packages/next/src/components/lightbox-overlay.tsx
```diff
@@ -0,0 +1,86 @@
+import type { ReactNode } from 'react';
+
+/**
+ * Renders the lightbox overlay for displaying an enlarged image with interactions.
+ *
```
justlevine, line 5: Drop a `@see {link}` ref to the gb html we're mimicking

packages/next/src/components/lightbox-overlay.tsx
```diff
@@ -0,0 +1,86 @@
...
+ * @return The lightbox overlay.
+ */
+export function LightboxOverlay(): ReactNode {
+	const closeButtonLabel = 'Close';
+	const closeButtonColor = 'var(--wp--preset--color--contrast)';
+	const backgroundColor = 'var(--wp--preset--color--base)';
```
justlevine, line 11: Add a comment explaining what these are, why theyre hardcoded, and where they're coming from.

packages/next/src/index.ts
ayushnirwal, line 1: `LightboxOverlay`, unlike the other components in this directory, is neither a wrapper over a NextJS component nor a component that is supposed to work only in a NextJS environment. 
Since we don't currently have a package for React components, I would say colocating with the template renderer makes more sense.

packages/next/src/template-renderer/index.tsx
```diff
@@ -53,6 +54,7 @@ export async function TemplateRenderer( {
 				</main>
 			</TemplateScripts>
 			{ /* Hot Fix for adding classes to the body outside the root layout */ }
```
ayushnirwal, line 56: Add a todo to check if we can patch this with either the portal or the `@wordpress/hooks` implementation.

ayushnirwal (changes requested):

ayushnirwal (approved):

packages/next/src/template-renderer/index.tsx
justlevine: Whats this hot fix comment refer to?

packages/next/src/template-renderer/index.tsx
```diff
@@ -54,6 +55,9 @@ export async function TemplateRenderer( {
...
 
+			{ /* @todo: Check if we can patch this with either the portal or the @wordpress/hooks implementation. */ }
+			<LightboxOverlay />
```
justlevine, line 59: Put this (todo included) in `<TemplateFooter>`, that way it's just the footer "portal" that will need to change and not our `TemplateRenderer`.

packages/blocks/src/blocks/core-image.tsx
```diff
@@ -248,17 +249,13 @@ const getImageProps = (
  *
  * @return Whether the lightbox is enabled.
  */
-//@ts-ignore -- Stubbed until lightbox support is fixed.
 const isLightboxEnabled = ( lightbox?: string | null ): boolean => {
```
justlevine, line 252: Update the doc-block to explicity note that if enabled the `<LightboxOverlay>` component will added in the template render, and a `@todo` to colocate the overlay here and portal it where needed

justlevine (changes requested):
Last nits, and this is IMO good to merge.

.changeset/legal-laws-walk.md
```diff
@@ -0,0 +1,6 @@
+---
+"@snapwp/blocks": patch
+"@snapwp/next": patch
+---
+
+fix: Images not opening in lightbox
```
justlevine, line 6:
```suggestion
fix: Add shim to support Image Block lightboxes
```

packages/next/src/template-renderer/index.tsx
Swanand01: We don't have a `<TemplateFooter>`. Do you mean putting this inside `<TemplateScripts>`?

justlevine: No I meant to create a new wrapper component that matches the DX and naming conventions of the existing component.

That component will hold our manual injections like this lightbox until we can turn it into a portal

packages/blocks/src/blocks/core-image.tsx
```diff
@@ -242,23 +243,20 @@ const getImageProps = (
...
-	// return lightboxObj.enabled;
-
-	// Disabling lightbox support for now.
-	return false;
+	if ( ! lightbox ) {
+		return false;
+	}
+	const lightboxObj = JSON.parse( lightbox ) as LightBoxProp;
+
+	return lightboxObj.enabled;
```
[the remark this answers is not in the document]
justlevine: This is valid feedback, don't trust GraphQL to give us a valid string.

PR #165, opened 2025-04-23
fix(ci): use locally built version of snapwp starter template

L0RD-ZER0:
## What
Change the use of `npm link` in ci with a temporary local install.

## Why
CLI version used after `npm link` is still the one from npm registry, and not the one built locally.

- https://github.com/rtCamp/headless/issues/460

## How
Replaced use of `npm link` in ci with a local install for locally built CLI version and then use it via `npx` to avoid using global version of SnapWP CLI. During internal testing, this turned out to be less error prone, whereas for `npm link`, it would sometimes use local version and sometimes one from npm registry, and show different errors randomly.

## Checklist

### DIFF
--- a/.github/workflows/pull_request_examples.yml
+++ b/.github/workflows/pull_request_examples.yml
@@ -118,6 +118,7 @@ jobs:
               run: |
                   cp '${{ github.workspace }}/examples/${{ matrix.example }}/.env.example' .env
                   sed -i 's|NEXT_PUBLIC_WP_HOME_URL=.*|NEXT_PUBLIC_WP_HOME_URL=http://localhost|' .env
+                  # Had to resort to a local install because npm link didn't want to work properly. Refer #165
                   npm i '${{ github.workspace }}/packages/cli' --no-save
                   npx snapwp --proxy <<< '.'
                   npm i
--- a/.github/workflows/pull_request_examples.yml
+++ b/.github/workflows/pull_request_examples.yml
@@ -100,7 +100,6 @@ jobs:
                   npm ci --workspaces --include-workspace-root
                   npm run --if-present local-registry:start
                   npm run --if-present build
-                  npm link snapwp
 
                   # Ensure Verdaccio has the right permissions for CI
                   touch .verdaccio/conf/htpasswd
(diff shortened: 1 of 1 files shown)


.github/workflows/pull_request_examples.yml
```diff
@@ -120,7 +119,7 @@ jobs:
                   cp '${{ github.workspace }}/examples/${{ matrix.example }}/.env.example' .env
                   sed -i 's|NEXT_PUBLIC_WP_HOME_URL=.*|NEXT_PUBLIC_WP_HOME_URL=http://localhost|' .env
                   npm i '${{ github.workspace }}/packages/cli' --no-save
```
justlevine, line 121: (assuming @ayushnirwal approves these deviations):

We need comments here and anywhere else we're deviating from `DEVELOPMENT.md`, i.e. explaining why we're using `npm install` + `npx snapwp` instead of `npm link` + `snapwp`

ayushnirwal:
@L0RD-ZER0 You are now using linking, you are outright installing the CLI package from the file system, which is one of the ways of using npm packages. ( others are linking and registry proxy ).

Can you educate me on why it does not work in CI? Are there any open issues about this already? This will help me in the future when I’m deciding between installing, linking, and registry proxy for sharing packages between JS projects.

Add that details as inline comments too.

ayushnirwal:
Would you mind trying [`npm_config_registry`](https://docs.npmjs.com/cli/v10/using-npm/registry) rather than installing the `snapwp` command?

```bash
npm_config_registry="<verdaccio-url>" npx snapwp --proxy
``` 

If that works, we would only be relying on the proxied registry for sharing dependencies. 

@justlevine I don't think we should change the usage docs because of this change, since this is a CI environment issue. Linking is faster than publishing to the local registry and provides a better experience.

L0RD-ZER0:
When we change the scaffolding flow to the following:

```bash
  cp '/home/runner/work/snapwp/snapwp/examples/nextjs/starter/.env.example' .env
  sed -i 's|NEXT_PUBLIC_WP_HOME_URL=.*|NEXT_PUBLIC_WP_HOME_URL=http://localhost|' .env
  npm_config_registry="http://localhost:487[3](https://github.com/L0RD-ZER0/snapwp/actions/runs/14638778171/job/41076108582#step:7:3)" npx snapwp --proxy
  npm i
```

We get the following error log during the `npx snapwp --proxy` step:

```
Thanks for using SnapWP!

Where would you like to create your new Headless WordPress frontend?
Please enter a relative or absolute path: [default: ./snapwp-app]
npm error code ENOENT
npm error syscall open
npm error path /home/runner/work/_temp/snapwp-test-app/package.json
npm error errno -2
npm error enoent Could not read package.json: Error: ENOENT: no such file or directory, open '/home/runner/work/_temp/snapwp-test-app/package.json'
npm error enoent This is related to npm not being able to find a file.
npm error enoent
```

@ayushnirwal @justlevine

ayushnirwal (approved):

justlevine (approved):

PR #167, opened 2025-04-28
docs: Documentation for Overloading Query Engines.

Ta5r:
## What

- This PR updates the documentation to guide users on creating custom query engines and configuring them in snapwp.config.ts, including how to switch between Apollo Client and TanStack Query.

## Why

- It enables users to easily implement and configure custom query engines, providing flexibility in choosing or swapping between different GraphQL clients.

### Related Issue(s):

- Closes [Task: Document usage of Query Engine and Packages](https://github.com/rtCamp/headless/issues/462)

## How

- The PR adds clear instructions on implementing custom QueryEngine classes, configuring engines in snapwp.config.ts, and switching between query engines.
- Added `query-enigne.md` documentation.
- Updated plugin README.md files with installation + usage instructions
- Updated Config API docs to document new properties + usage.
- Updated any other README.md files with new/changes to already-documented components.

## Checklist


docs/config-api.md
```diff
@@ -53,13 +60,47 @@ export default config;
...
-| `blockDefinitions` | `BlockDefinitions`       | [blocks](../packages/blocks/src/blocks/index.ts)                | Block definitions for the editor.<br />[Learn more](./overloading-wp-behavior.md#overloading-blocks)                                            |
+| Property           | Type                     | Default Value                                                              | Description                                                                                                                                     |
+| ------------------ | ------------------------ | -------------------------------------------------------------------------- | ----------------------------------------------------------------------------------------------------------------------------------------------- |
+| `query`            | `QueryEngine`            | [ApolloClientEngine](../packages/plugin-apollo-client/src/engine/index.ts) | Configuration for the GraphQL query engine.<br />See below for more details on how to customize it.[Learn more](./query-engine.md)              |
```
justlevine, line 65: It's not a default value if you manually need to define it in your config.

docs/config-api.md
```diff
@@ -53,13 +60,47 @@ export default config;
...
-| ------------------ | ------------------------ | --------------------------------------------------------------- | ----------------------------------------------------------------------------------------------------------------------------------------------- |
-| `parserOptions`    | `HTMLReactParserOptions` | [defaultOptions](../packages/next/src/react-parser/options.tsx) | The default options for the `html-react-parser` library.<br />[Learn more](./overloading-wp-behavior.md#2-pass-customparseroptions-to-overload) |
-| `blockDefinitions` | `BlockDefinitions`       | [blocks](../packages/blocks/src/blocks/index.ts)                | Block definitions for the editor.<br />[Learn more](./overloading-wp-behavior.md#overloading-blocks)                                            |
+| Property           | Type                     | Default Value                                                              | Description                                                                                                                                     |
```
justlevine, line 63: All of these properties, and the subsections should be sorted AZ now that there's more than 2.

packages/plugin-apollo-client/README.md
```diff
@@ -10,6 +10,39 @@ SnapWP's query system and configuration API are built to support any GraphQL cli
 > 🐉 There be dragons!
 > This project is in **active development** and considered _experimental_. Some features may be incomplete, unstable, or subject to change.
 
+## Installation
+
+To integrate plugin-apollo-client within the SnapWP framework:
+
+```bash
+npm install plugin-apollo-client
```
justlevine, line 18:
```suggestion
npm install plugin-apollo-client --save
```

packages/plugin-apollo-client/README.md
```diff
@@ -10,6 +10,39 @@ SnapWP's query system and configuration API are built to support any GraphQL cli
...
+const config: SnapWPConfig = {
+	query: {
+		engine: ApolloClientEngine,
+	},
+};
+
+export default config;
+```
+
+#### Breakdown of the Setup
```
justlevine, line 38: Instead of having separate sections, the entire `## Installation ` section should be numbered steps with e.g

----
1.  Add the package to your project's dependencies:

    ```console
    npm install --save @snapwp/plugin-apollo-client
    ```
    
2. Update your `snapwp.config.ts` to use the Query Engine:

    ```diff
    import type { SnapWPConfig } from '@snapwp/core/config';
    + import { ApolloClientEngine } from '@snapwp/plugin-apollo-client';

    const config: SnapWPConfig = {
    +	query: {
    +		engine: ApolloClientEngine,
    +	},
    };

    export default config;
    ```

justlevine: (PS: do we need/want to add a note in step one about installing `@apollo/client` directly too? (or tanstack/query for that readme)

packages/core/README.md
```diff
@@ -51,6 +51,32 @@ import { getConfig } from '@snapwp/core/config';
 const config = getConfig();
 ```
 
+#### Changing Query Engine from Apollo Client (default) to TanStack Query
```
justlevine, line 54: If this section belongs anywhere, it's in the config api docs, not this core readme that just shows the method signature for using `getConfig()`

docs/config-api.md
Ta5r: We are shipping the `examples/starter` with the `snapwp.config.ts` file already using the `ApolloClientEngine` hence I think this makes it the **default**. Confirmed it with @SH4LIN  🙇 
Please let me know if there is something i am misinterpreting 🙇

docs/config-api.md
Ta5r: Done ✅ .

packages/plugin-apollo-client/README.md
Ta5r: Done ✅ for both packages.

packages/core/README.md
Ta5r: Done ✅ .

SH4LIN (approved):
LGTM!

docs/config-api.md
justlevine: What you're missing is that we're documenting the _framework_ in this file, not `examples/nextjs/starter`. In context, `default` refers to the _default value_ of the configuration object ( i.e. `SnapWPConfig.query.engine` if left unset ), and not what behavior we've chosen to illustrate.

docs/query-engine.md
```diff
@@ -0,0 +1,256 @@
...
+
+	useQuery< TData, TQueryVars >( {
+		name,
+		query,
+		options,
+	}: QueryArgs< TData, TQueryVars > ): TData {
+		// Implement your client-side fetching here.
+	}
+
+	QueryProvider = QueryProvider;
```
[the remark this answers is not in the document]
justlevine: @SH4LIN, @ayushnirwal is this correct?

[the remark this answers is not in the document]
SH4LIN: Technically, QueryProvider is a functional component adding queryProviderInstance will make it more confusing between functional component and class component, I don't think this will be circular as we are importing `QueryProvider`.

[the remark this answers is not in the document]
SH4LIN: @justlevine Initially it was QueryProvider = ApolloQueryProvider/TanstackQueryProvider, we changed it after your feedback to refactor ApolloQueryProvider to QueryProvider. Technically it doesn't have any problems so I changed it, but I don't have any explanation for QueryProvider = QueryProvider.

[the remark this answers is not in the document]
justlevine: @SH4LIN confirm plz: https://github.com/rtCamp/snapwp/pull/167/commits/b626f3a76213d1fae4777f009eec40d80682497b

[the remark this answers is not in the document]
justlevine: Yup that did _not_ go as expected: https://github.com/rtCamp/snapwp/actions/runs/14879567675/job/41784345415?pr=167 

What am I missing?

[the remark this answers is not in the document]
SH4LIN: @justlevine Left hand side is what we care about so it should be there, we don't care about the name on the right hand side it can be anything. You removing entire class property broke it. Assignment is required, the one in question was x=x (this x=x was only in name and not actual x functionality = x functionality)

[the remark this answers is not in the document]
justlevine: @SH4LIN since you seem to have a grasp of what is required please do whatever it is your suggesting as I still don't understand what you're trying to convey, and it's likely less work than the amount of back-and-forth we've done on this 😉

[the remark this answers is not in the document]
ayushnirwal: >What's confusing to me is the X = itself syntax.

@justlevine it's not `X = itself`. This weirdness is caused because QueryProvider needs "use client" directive since it's a client-side component. Therefore, we can't define the method in the class definition itself. So, it is defined in a new file, and the reference is passed as a method to the class.

In this case, the method name in the interface and the function reference are coincidentally named same, causing the `x=x` line.  We can solve the confusion by renaming the import (function reference) `QueryProvider` as `ApolloQueryProvider`.

> The assignment 'QueryProvider = QueryProvider;' is circular and may confuse readers.

@SH4LIN Yes, I agree this will confuse the reader. We should import `QueryProvider` as `ApolloQueryProvider`. The method `QueryProvider` of class `ApolloQueryEngine` will get the function reference of `QueryProvider` in the file `packages/plugin-apollo-client/src/engine/query-provider`.

Same goes for the tanstack plugin too.

justlevine (approved):

PR #168, opened 2025-04-29
Feature/static exports

SH4LIN:
## What

- This PR adds the `generateStaticParams()` function to the Next.js app, which uses functions from the `@snapwp/next` and `@snapwp/query` packages to fetch paths from the WordPress server and statically generate them at build time.  
- We can export this statically by setting `output: 'export'` in the Next.js config. However, we are not including it by default, as it is up to the developer’s preference. Pages will still be statically generated in the Next.js app without it.

### Related Issue(s):

- Closes: https://github.com/rtCamp/headless/issues/439

## How

- Fetching 100 pages, 100 posts and 100 terms URI from WP server using GraphQL request and passing it to the generateStaticParams() function which will statically generate pages during build.

## Testing Instructions

1. cd `examples/nextjs/starter`
2. npm run build
3. npm run start

## Checklist


packages/query/src/queries/get-pages-to-render-statically.graphql
justlevine, line 1: Probably will look something closer to this once we're done:
```graphql
fragment FowardPaginationInfo on WPPageInfo {
  hasNextPage
  endCursor
}

fragment PathInfo on UniformResourceIdentifiable {
  uri
  id
}

query MultiPagination(
  $first: Int = 100,
  $hasMoreContentNodes: Boolean = true, $contentNodeCursor: String = null,
  $hasMoreTerms: Boolean = true, $termCursor: String = null,
  $hasMoreUsers: Boolean = true, $userCursor: String = null
) {
  contentNodes(first: $first, after: $contentNodeCursor) @skip(if: $hasMoreContentNodes) {
    pageInfo {
      ...FowardPaginationInfo
    }
    nodes {
      ...PathInfo
    }
  }
  terms(first: $first, after: $termCursor) @skip(if: $hasMoreTerms) {
    pageInfo {
      ...FowardPaginationInfo
    }
    nodes {
      ...PathInfo
    }
  }
  users(first: $first, after: $userCursor) @skip(if: $hasMoreUsers) {
    pageInfo {
      ...FowardPaginationInfo
    }
    nodes {
      ...PathInfo
    }
  }
}
````

**Review by justlevine — CHANGES_REQUESTED** [review]
Looks good from a quick pass. Need @ayushnirwal to do a real review

File: packages/next/src/static-exports/index.ts [hunk 1]
```diff
@@ -0,0 +1,43 @@
+import { QueryEngine } from '@snapwp/query';
+
+/**
+ * Get the paths to render statically.
+ *
+ * @return The paths to render statically.
+ */
+export const getPathsToRenderStatically = async (): Promise<
```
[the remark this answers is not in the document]
SH4LIN: Hi @justlevine,
In the Next.js Pages Router, dynamic pages require exporting a function named getStaticPaths:
https://nextjs.org/docs/pages/building-your-application/data-fetching/get-static-paths

I'm not opposed to converting this into getStaticPaths(), but just wanted to check—do you think this might be a bit confusing for others?

packages/next/src/static-exports/index.ts
```diff
@@ -0,0 +1,43 @@
...
+> => {
+	const pathsToRender = await QueryEngine.getPaths();
+
+	const paths: Array< { slug: string[] } > = [];
+
+	if ( pathsToRender.pages ) {
+		for ( const page of pathsToRender.pages ) {
+			if ( page.uri ) {
+				const slug = page.uri.split( '/' ).filter( Boolean );
+				paths.push( { slug } );
```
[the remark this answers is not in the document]
SH4LIN: We want uri and we are using uri `page.uri.split( '/' ).filter( Boolean )` but after seo PR our catch all paths variable name converted into slug i.e [[...slug]], so we have to create an array of object with slug property init.

packages/next/src/static-exports/index.ts
```diff
@@ -0,0 +1,43 @@
...
+
+/**
+ * Get the paths to render statically.
+ *
+ * @return The paths to render statically.
+ */
+export const getPathsToRenderStatically = async (): Promise<
+	Array< { slug: string[] } >
+> => {
+	const pathsToRender = await QueryEngine.getPaths();
```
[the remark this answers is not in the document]
SH4LIN: Do you have any specific name in mind?

packages/next/src/static-exports/index.ts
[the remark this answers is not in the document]
justlevine: I'm not sure. @ayushnirwal ?

packages/query/src/queries/get-pages-to-render-statically.graphql
```diff
@@ -0,0 +1,38 @@
...
+			id
+		}
+	}
+
+	terms(first: $first, after: $termCursor) @include(if: $hasMoreTerms) {
+		pageInfo {
+        	hasNextPage
+        	endCursor
+        }
+        nodes {
```
justlevine, line 23: formatting

packages/query/src/query-engine/index.ts
justlevine, line 1: Sort this file plz:

1. Import order (also cleanup the `import type { } from '@snapwp/core`)
2. TS Types. Local, then Exportable. 
3. _Then_ the `export class QueryEngine {}`

packages/next/src/static-exports/index.ts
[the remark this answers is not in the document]
justlevine: Not sure I follow. Using `mysite/about/philosophy` as an example `Page.uri` :

- the _`slug`_ is `philosophy`
- our `templateByUri` takes the `/about/philosophy` **URI**
- the `[[...slug]]` catch-all in the `/example` is _also_ `/about/philosophy`, despite being called `slug` for some reason.

packages/next/src/middleware/utils.ts
```diff
@@ -1,7 +1,6 @@
 import { NextResponse, type NextMiddleware } from 'next/server';
 import { getConfig } from '@snapwp/core/config';
 import { corsProxyMiddleware } from './cors';
-import { currentPath as cm } from './current-path';
```
justlevine, line 4: Feel like we should deprecate this instead of deleting.

examples/nextjs/starter/src/app/[[...slug]]/page.tsx
```diff
@@ -19,12 +19,19 @@ export async function generateMetadata( {
...
-		<TemplateRenderer>
+		<TemplateRenderer params={ params }>
 			{ ( editorBlocks ) => {
 				return <EditorBlocksRenderer editorBlocks={ editorBlocks } />;
 			} }
 		</TemplateRenderer>
 	);
 }
+
+export const generateStaticParams = async () => {
```
Comment (justlevine) [inline examples/nextjs/starter/src/app/[[...slug]]/page.tsx:32]: Move the other `generate*` down here too.

justlevine (changes requested):
@SH4LIN If you need help with GraphQL pagination lmk

packages/query/src/query-engine/index.ts
SH4LIN: Imports are sorted by linters automatically after adding `import type { } from '@snapwp/core`.

Eslint is configured as below
```
'builtin',
'external',
'internal',
'parent',
'sibling',
'index',
'type',
```

packages/next/src/middleware/utils.ts
SH4LIN: We are planning to create new example for this, so for existing example/nextjs/starter this will be required, may need to add it again. @justlevine Should I add it back again now or in the follow up task?

justlevine: not fully sure I follow, but either way sounds like you should _delete_ existing exports from `current-path.ts` file, and at most update the documentation there if it's currently unused.

packages/query/src/query-engine/index.ts
justlevine: Great that's No. 1, now move the types from the bottom of this file up before the class export.

packages/next/src/static-exports/index.ts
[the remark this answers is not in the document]
SH4LIN: So, `[[...slug]]` is a name we choose. Whenever we access a URL like `mysite/about/philosophy`, the value of `slug` becomes:
`{ slug: ['about', 'philosophy'] }`

We can name it anything, such as `[[...x]]`. For example, if we used `[[...path]]`, the value would be:
`{ path: ['about', 'philosophy'] }`

Since we’ve used `slug` here, we need to provide an array of objects like this:
`[ { slug: ['about', 'philosophy'] } ]`

This is the same structure we receive from `params`. Next.js will use this array of objects to determine the paths to statically generate on the server.

packages/query/src/queries/get-pages-to-render-statically.graphql
```diff
@@ -0,0 +1,38 @@
...
+ 	$first: Int = 100,
+ 	$hasMoreContentNodes: Boolean = false, $contentNodeCursor: String = null,
+	$hasMoreTerms: Boolean = false, $termCursor: String = null,
+	$hasMoreUsers: Boolean = false, $userCursor: String = null
+) {
+	contentNodes(first: $first, after: $contentNodeCursor) @include(if: $hasMoreContentNodes) {
+		pageInfo {
+			hasNextPage
+			endCursor
+		}
```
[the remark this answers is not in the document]
SH4LIN: When fragments were used it was not able to identify keys properly.

image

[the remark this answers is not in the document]
justlevine: Okay, once you've handled the other comments on that file (e.g. https://github.com/rtCamp/snapwp/pull/168/files#r2075121605 ), I'll push a commit (and walk you through how ) to [unmask the fragments](https://[the-guild.dev/graphql/codegen/plugins/presets/preset-client#fragment-masking-with-nested-fragments](https://the-guild.dev/graphql/codegen/plugins/presets/preset-client#fragment-masking-with-nested-fragments))

packages/query/src/query-engine/index.ts
justlevine: done in https://github.com/rtCamp/snapwp/pull/168/commits/061038ef4f7d4193c55a9410cac6db8e7579d43e

packages/query/src/queries/get-pages-to-render-statically.graphql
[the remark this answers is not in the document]
justlevine: Resolved in https://github.com/rtCamp/snapwp/pull/168/commits/bc0bc8c7ad98cf2c7af77e65f319b12c053b02f2

packages/query/src/query-engine/index.ts
```diff
@@ -62,4 +65,102 @@ export class QueryEngine {
...
+			hasMoreContentNodes,
+			hasMoreTerms,
+			hasMoreUsers,
+			contentNodeCursor,
+			termCursor,
+			userCursor,
+		};
+
+		do {
+			const data: GetPagesToRenderStaticallyQuery = await fetchQuery( {
```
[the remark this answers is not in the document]
justlevine: https://github.com/rtCamp/snapwp/pull/168/commits/bc0bc8c7ad98cf2c7af77e65f319b12c053b02f2

justlevine (changes requested):
@SH4LIN PTAL at bc0bc8c7ad98cf2c7af77e65f319b12c053b02f2 and specifically the refactored `QueryEngine.getStaticPaths()` function - as I may have accidently introduced some regressions.

Other than than [this still-open thread](https://github.com/rtCamp/snapwp/pull/168/files#r2072696894) my only remaining question is why we even need the function to be _inside_ QueryEngine, instead of just using QueryEngine.fetchQuery() inside of `static-exports/index.ts` directly?

cc @ayushnirwal

packages/query/src/query-engine/index.ts
```diff
@@ -62,4 +79,84 @@ export class QueryEngine {
...
+	): Promise< Record< string, PathInfo[] > > => {
+		const data = await fetchQuery( {
+			name: 'GetPagesToRenderStatically',
+			query: GetPagesToRenderStaticallyDocument,
+			options: {
+				variables: {
+					...variables,
+					hasMoreContentNodes: !! variables.contentNodesCursor,
+					hasMoreTerms: !! variables.termsCursor,
+					hasMoreUsers: !! variables.usersCursor,
```
SH4LIN, line 105: @justlevine I think we should not keep this? As we are trying to create dynamic indeterminate loop on bases of the sections available in the query, this defeats it's purpose, instead we should just destructure ...variables here. And when creating newVariables we should add hasMore{key} key.

justlevine: Agreed, but i think pause on the refactoring until I hear a good reason why not to move this entire function into `static-exports/index.ts`

packages/query/src/query-engine/index.ts
```diff
@@ -62,4 +79,84 @@ export class QueryEngine {
...
+			throw new Error(
+				'Error fetching pages to render statically. No data returned.'
+			);
+		}
+
+		let resolvedData: Record< string, PathInfo[] > = {};
+
+		let shouldFetchMore = false;
+
+		const newVariables: Record< string, string > = {};
```
SH4LIN, line 120: making this string, string we are removing it's guard rail

packages/query/src/query-engine/index.ts
SH4LIN: The main reason is that all our request-making functions are currently defined within this class, and it’s generally a good practice to colocate functions that serve similar purposes. This also sets us up for potential future enhancements—such as introducing function overloading like we did with the query engine—by keeping related logic together.

justlevine: Yeah , sorry I don't buy that 

1. None of those functions are _related_ to each other, they just use `fetchQuery()`.
2. The two existing functions are _only_ there as a matter of tech debt. However, they at least "server similar purposes" in that theyre a rendering requirement. This is a build or a sitemap requirement, and honestly id move those other functions off this class in a future PR.
3. Future enhancements argument is a false premise. See `seo-manager.tsx` if you disagree.

SH4LIN: > None of those functions are related to each other.

I wasn’t suggesting that these functions are related to each other directly, but rather that they serve a similar purpose — fetching data or making GraphQL requests. That said, if you still feel it's more appropriate, I’m happy to move this function into the next package.

packages/query/src/query-engine/index.ts
```diff
@@ -62,4 +79,84 @@ export class QueryEngine {
...
+			},
+		} );
+
+		if ( ! data ) {
+			throw new Error(
+				'Error fetching pages to render statically. No data returned.'
+			);
+		}
+
+		let resolvedData: Record< string, PathInfo[] > = {};
```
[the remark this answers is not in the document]
SH4LIN: Yes, we are not using ID anywhere we can change it from pathInfo[] to string[]

packages/query/src/query-engine/index.ts
justlevine: Confirm w. @ayushnirwal , but yes [I believe](https://github.com/rtCamp/snapwp/pull/168#pullrequestreview-2822891387) that's where it's best suited.

packages/query/src/query-engine/index.ts
SH4LIN: @justlevine Thank you for the idea of using satisfies, I made slight change to the code, instead of using satisfies on the final object, we can use the satisfies key word on hasMore key and cursor key.

```JS
const cursorKey = `${ resolvedKey }Cursor` satisfies keyof VariableTypes;
const hasMoreKey = `hasMore${ capitalize(
	resolvedKey
) }` satisfies keyof VariableTypes;
if ( pageInfo.hasNextPage && pageInfo.endCursor ) {
	shouldFetchMore = true;
	newVariables[ cursorKey ] = pageInfo.endCursor;
	newVariables[ hasMoreKey ] = true;
} else {
	newVariables[ cursorKey ] = null;
	newVariables[ hasMoreKey ] = false;
}
```

packages/query/src/query-engine/index.ts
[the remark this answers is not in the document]
SH4LIN: Will it be okay to return only string[] array instead of Record< keyof GetPagesToRenderStaticallyQuery, string[] >?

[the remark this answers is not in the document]
justlevine: If you think that's the ideal shape go for it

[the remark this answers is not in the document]
SH4LIN: Follow up question, do you think now or in future we will need keyof GetPagesToRenderStaticallyQuery in any way?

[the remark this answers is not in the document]
justlevine: Definitely not if this gets moved into `static-exports`.
If part of this is staying generic (e.g. to DRY out sitemaps), then _maybe_ but we can already refactor in the future if so.

packages/query/src/query-engine/index.ts
```diff
@@ -62,4 +65,95 @@ export class QueryEngine {
...
 	};
+
+	/**
+	 * Fetches paths to be rendered statically.
+	 *
+	 * @param {GetPagesToRenderStaticallyQueryVariables} variables - The variables for the query.
+	 * @return The path data to be rendered statically.
+	 */
+	static getStaticPaths = async (
+		variables: GetPagesToRenderStaticallyQueryVariables = {}
```
justlevine, line 76: Dont bring these types directly into our class, it causes coupling with GraphQL we want to avoid.

Instead our VariableType is validated as compatible with the graphql via `fetchQuery()`

packages/query/src/query-engine/index.ts
```diff
@@ -62,4 +65,95 @@ export class QueryEngine {
...
+
+			resolvedData.push(
+				...nodes
+					.filter( ( { uri } ) => !! uri )
+					.map( ( { uri } ) => uri! )
+			);
+
+			const cursorKey =
+				`${ resolvedKey }Cursor` satisfies keyof GetPagesToRenderStaticallyQueryVariables;
+			const hasMoreKey = `hasMore${ capitalize(
```
justlevine, line 121: Didn't I get rid of this needless `capitalize()` by inverting the query variable name?

packages/query/src/query-engine/index.ts
```diff
@@ -62,4 +65,95 @@ export class QueryEngine {
...
+		if ( ! data ) {
+			throw new Error(
+				'Error fetching pages to render statically. No data returned.'
+			);
+		}
+
+		const resolvedData: string[] = [];
+		let shouldFetchMore = false;
+		const newVariables = {
+			...variables,
```
justlevine, line 97: Why are you dirtying up the state

packages/query/src/query-engine/index.ts
```diff
@@ -62,4 +65,95 @@ export class QueryEngine {
...
+
+			if ( pageInfo.hasNextPage && pageInfo.endCursor ) {
+				shouldFetchMore = true;
+				newVariables[ cursorKey ] = pageInfo.endCursor;
+				newVariables[ hasMoreKey ] = true;
+			} else {
+				newVariables[ cursorKey ] = null;
+				newVariables[ hasMoreKey ] = false;
+			}
+		} );
```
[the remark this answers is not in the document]
SH4LIN: We need to explicitly set it to false otherwise if it is unset then GraphQL will take the default value for hasMoreKey = true, and it will make the request for part which does not have more data.

packages/query/src/query-engine/index.ts
```diff
@@ -62,4 +79,77 @@ export class QueryEngine {
...
+
+		// Recursively fetch more data if there are more pages to fetch.
+		if ( shouldFetchMore ) {
+			const additionalData = await QueryEngine.getStaticPaths( {
+				first: 100,
+				...newVariables,
+			} );
+
+			resolvedData.push( ...additionalData );
+		}
```
justlevine, line 151: Changing the function like so + updating the GQL `*HasMore` variables to default to `true`, will allow us to call await `QueryEngine.getStaticPaths( { first: 100, } );` i.e. w/o the explicit node references.

```suggestion
			const { nodes, pageInfo } = section as {
				nodes: StaticRouteNodeDataFragment[];
				pageInfo: StaticRoutePageInfoFragment;
			};

			// Store the data.
			resolvedData.push(
				...nodes
					.filter( ( { uri } ) => !! uri )
					.map( ( { uri } ) => uri! )
			);

			// Set the pagination variables for the next fetch.
			const cursorKey =
				`${ resolvedKey }Cursor` satisfies keyof VariableType;
			const hasMoreKey =
				`${ resolvedKey }HasMore` satisfies keyof VariableType;

			if ( pageInfo.hasNextPage && pageInfo.endCursor ) {
				shouldFetchMore = true;
				newVariables[ cursorKey ] = pageInfo.endCursor;
			}
			newVariables[ hasMoreKey ] = !! newVariables?.[ cursorKey ];
		} );

		// Recursively fetch more data if there are more pages to fetch.
		if ( shouldFetchMore ) {
			const additionalData = await QueryEngine.getStaticPaths( {
				first: 100,
				...newVariables,
			} );

			resolvedData.push( ...additionalData );
		}
```

justlevine: Not sure why that diff was weird. Other than the inline comments (add some!) the important part is moving `newVariables[ hasMoreKey ] = !! newVariables?.[ cursorKey ];` _outside_ of the conditional.

justlevine (commented):
Only thing left is to get @ayushnirwal 's feedback on the open conversations, then docs and a changeset and we should be good to merge.

ayushnirwal (changes requested):

ayushnirwal:
@SH4LIN Can you put the problem with statically building pages here? I'm talking about NextJS DOS'ing the WP server while building.

SH4LIN:
> @SH4LIN Can you put the problem with statically building pages here? I'm talking about NextJS DOS'ing the WP server while building.

@ayushnirwal @justlevine StaticExports are generated using a workers where NextJS uses 4 workers as fallback values, when these workers sends out request after request to WP server simultaneously WP server hangs up and starts throwing gateway timeout error.

One workaround for this is to create queue mechanism but NextJS doesn't any direct mechanism to control it. So I started reviewing NextJS code.
First solution that I found was
NexJS provides a config which allows to set the CPUS config values using which nextJS decides how many workers to use. So if we reduce the number of CPUS to 1 it sets the number of workers to be 1 as well. `config.experimentals.cpus = 1`.
Second solution that I found was to set following experimental configs.
```
// Control how many pages will be processed per worker
staticGenerationMaxConcurrency: 2,
// Set minimum pages before spinning up a new worker
staticGenerationMinPagesPerWorker: 25,
// How many retry attempts for failed page generation
staticGenerationRetryCount: 1
```

packages/query/src/query-engine/index.ts
SH4LIN: @justlevine If we create dynamic queries then we won't know the variables type, how should I handle that?

justlevine:
> StaticExports are generated using a workers where NextJS uses 4 workers as fallback values, when these workers sends out request after request to WP server simultaneously WP server hangs up and starts throwing gateway timeout error.

Is the issue about _concurrency_, or just the weight of each query?

Other than making sure our queries are performant (and the user making sure they have their WP optimized), I think the way for us to handle this from a framework POV is to have a `config.perPage` so users can lower it from our default `100` to something that works better for their machine constraints.

packages/query/src/query-engine/index.ts
justlevine: @ayushnirwal the question here is whether there's any reason for a specific QueryEngine function, instead of just using `QueryEngine.fetchQuery(...)` directly in `'getWPStaticPaths' from './static-exports/index.ts'`

SH4LIN:
> Is the issue about concurrency, or just the weight of each query?

Both, WP server take lot of time to return data, and concurrency makes it more slow and causes timeout

packages/next/src/static-exports/index.ts
```diff
@@ -1,21 +1,99 @@
...
+			newVariables[ cursorKey ] = pageInfo.endCursor;
+		}
+		newVariables[ hasMoreKey ] = !! newVariables?.[ cursorKey ];
 	} );
 
-	return paths;
+	// Recursively fetch more data if there are more pages to fetch.
+	if ( shouldFetchMore ) {
+		const additionalData = await getWPStaticPaths( {
+			first: 100,
```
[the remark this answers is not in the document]
SH4LIN: Yes, also we need to add flag whether they want paginated results or not. @justlevine @ayushnirwal What are your input on that?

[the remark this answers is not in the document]
justlevine: Why would a user not want us to paginate?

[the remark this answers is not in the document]
SH4LIN: If pagination is enabled, reducing the first variable will not have a significant impact. For instance, if first is set to 10, the pagination logic will still result in fetching all posts—albeit in batches of 10 per page—until the complete dataset is retrieved. The performance issue is with the complete dataset.

ayushnirwal:
> Is the issue about concurrency, or just the weight of each query?

Both. 

>Other than making sure our queries are performant (and the user making sure they have their WP optimized), I think the way for us to handle this from a framework POV is to have a config.perPage so users can lower it from our default 100 to something that works better for their machine constraints.

Since making the queries performant and the WP server optimized is out of the scope of this PR. Let's add the second solution in @SH4LIN 's [comment](https://github.com/rtCamp/snapwp/pull/168#issuecomment-2866429750) to the documentation as a proposed workaround. 
If a user experiences problems with the static build, they can use the workaround. IMO we should not rely on any (undocumented) experimental features in our framework's working.

ayushnirwal (approved):
Approving under the assumption that Docs will be added in a different PR.
Code LGTM.

packages/next/src/static-exports/index.ts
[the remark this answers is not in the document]
ayushnirwal: @justlevine TLDR: making this configurable does not solve the perf issue.

PR #176, opened 2025-05-07
chore: update NPM deps

justlevine:
## What

This PR updates our NPM deps by:

- Bumping/Adding support for latest versions
- Pinning snapwp package minimum versions
- Sorting dependencies/peerDeps
- `npm update`

As a result of the updates, the QueryEngine types had to be fixed.

## Why

`npm update` on it's own was causing a conflict because of `graphql-request`. Made sense to handle the rest at the same time.

## How

See diff.

## Checklist


packages/plugin-apollo-client/src/engine/index.ts
```diff
@@ -7,23 +7,22 @@ import {
...
-	type QueryOptions,
 	type ServerError,
 	type ServerParseError,
 } from '@apollo/client';
 import { Logger } from '@snapwp/core';
 import { getGraphqlUrl } from '@snapwp/core/config';
 import { QueryProvider as ApolloQueryProvider } from './query-provider';
 import type { TypedDocumentNode } from '@graphql-typed-document-node/core';
-import type { QueryEngine } from '@snapwp/types';
+import type { QueryEngine, QueryOptions } from '@snapwp/types';
```
justlevine, line 17: This (versus the `@apollo/client.QueryOptions`) is what primarily fixed the ts issues after fixing the interface to no longer be shorthand. Needs confirmation I didnt introduce a regression

packages/plugin-tanstack-query/package.json
```diff
@@ -28,11 +28,13 @@
 		"lint": "eslint . --config ../../.eslintrc.json",
 		"lint:fix": "eslint . --config ../../.eslintrc.json --fix"
 	},
+	"dependencies": {
+		"graphql-request": "^6.1.0 || ^7.1.2"
```
justlevine, line 32: - Not entirely sure why we need this at all, but the previous 7.1.2 conflicts with the one used internally by graphql-codegen.
- I _think_ this is a dependency (since its just a small implementation detail), but could use confirmation.

packages/types/package.json
```diff
@@ -31,13 +31,12 @@
...
 		"typecheck": "tsc --noEmit"
 	},
-	"dependencies": {
+	"peerDependencies": {
 		"react": "^19.0.0"
 	},
 	"devDependencies": {
 		"@types/node": "*",
 		"@types/react": "*",
-		"next": "^15",
```
justlevine, line 40: I didnt see us using this anywhere

packages/types/src/query-engine/index.ts
```diff
@@ -14,15 +14,15 @@ export interface QueryEngine< TClient > {
 	 *
 	 * @return The server-side query client instance.
 	 */
-	getClient(): TClient;
+	getClient: () => TClient;
```
justlevine, line 17: Changed per https://typescript-eslint.io/rules/method-signature-style/

packages/blocks/package.json
```diff
@@ -30,7 +30,11 @@
 		"typecheck": "tsc --noEmit"
 	},
 	"dependencies": {
-		"html-entities": "^2.5.2",
+		"html-entities": "^2.5.2"
+	},
+	"peerDependencies": {
+		"@snapwp/core": "^0.3.1",
```
ayushnirwal, line 36:
```suggestion
		"@snapwp/core": "*",
```

packages/plugin-apollo-client/src/engine/index.ts
ayushnirwal: After this change, `fetchQuery` takes options defined by type `@snapwp/types.QueryOptions` regardless of what plugin is installed and configured in the userland. 

@SH4LIN PTAL at this PR and advice on a real fix.

packages/plugin-tanstack-query/package.json
ayushnirwal: > Not entirely sure why we need this at all?

TRK does not directly support `graphQL` requests directly, it supports promises. `grqphql-request` is required to make promises for query and mutations.

>I think this is a dependency 
True.

ayushnirwal (changes requested):
This PR is almost complete in terms of dependency updates. I'm missing context on why types were needed to be updated.

justlevine:
> I'm missing context on why types were needed to be updated.

Because once the deps are "fixed", our [QueryEngine class lights up from typescript/eslint](https://github.com/rtCamp/snapwp/pull/176#discussion_r2078606602). I could just tsignore so it can be handled outside of this PR so we can merge this one.

packages/blocks/package.json
justlevine: Rationale for a specific number:

1. This is a peer dep, which means users don't have our root `package.json` to infer the values from.
2. This is a (probably invalid currently) _minimum_ version dependency. While I used the latest version for ease, this number will be blocked if/when `@blocks` requires a new minimum version of `@core`

packages/types/package.json
SH4LIN: @justlevine After the sitemap PR this dependency is added in the types package, again. Doing this will create unnecessary merge conflict. We need this in sitemap PR.

packages/plugin-apollo-client/src/engine/index.ts
justlevine: @ayushnirwal I've reverted the type changes in https://github.com/rtCamp/snapwp/pull/176/commits/e83a446059317d84a9648a76d01d6b59dff4ba93

@SH4LIN your task is now to remove those `eslint-disable`s I added to the interface, and then update this file to be compliant with the failing eslint rule: https://typescript-eslint.io/rules/method-signature-style/

ayushnirwal (approved):

packages/plugin-apollo-client/src/engine/index.ts
justlevine: We'll follow this up in a new pr if/as needed

PR #177, opened 2025-05-12
chore: add usage docs for seo

ayushnirwal:
## What
Adds documentation related to Seo systems.

### Related Issue(s):

https://github.com/rtCamp/headless/issues/431

## How
- Adds example showing how to register a plugin.
- Adds a section explaining requirements for a plugin to be registered.

## Checklist

### DIFF
--- a/docs/seo.md
+++ b/docs/seo.md
@@ -1,0 +1,55 @@
+# SEO Metadata
+
+SnapWP export 2 main functions `getPageMetadata` and `getLayoutMetadata` to add meta tags for SEO.
+`getLayoutMetadata` - Gets SEO data for the site. 
+`getPageMetadata` - Gets SEO data related to the template. Adds update icon 
+
+## Registering a new plugin
+The SEO functionality can be extended by registering new plugins. Below is an example of updated root layout.
+
+```typescript
+import { RootLayout } from '@snapwp/next';
+import { getLayoutMetadata, Seo } from '@snapwp/next/seo';
+import type { Metadata } from 'next';
+import type { PropsWithChildren } from 'react';
+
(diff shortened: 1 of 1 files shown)


docs/seo.md
```diff
@@ -0,0 +1,57 @@
...
+
+The SEO functionality can be extended by registering new plugins. Below is an example of updated root layout.
+
+```typescript
+import { RootLayout } from '@snapwp/next';
+import { getLayoutMetadata, Seo } from '@snapwp/next/seo';
+import type { Metadata } from 'next';
+import type { PropsWithChildren } from 'react';
+
+// Register a plugin before calling function `getLayoutMetadata`
```
justlevine, line 18: Registration should happen for now in RootLayout?

ayushnirwal: Yes, it's not connected to the config right now.

docs/seo.md
```diff
@@ -0,0 +1,57 @@
...
+}
+```
+
+## Anatomy of a SEO plugin
+
+An plugin requires 3 things to be registered.
+
+-   **Fragment** to define the data needed from the WP server.
+-   **Parser** to parse the fragment data to be consumable by NextJS.
+-   **Location** has value of either `page` or `layout`. This defines if the parsed meta data should be added to the returned value of `getPageMetadata` or `getLayoutMetadata`.
```
justlevine, line 54: I think we should provide the actual shape, if not an example of these. See the level of fidelity in query-engine.md

justlevine (changes requested):
A bit more on what needs to be provided to registerPlugin, and we should be good to merge.

